PR #3, opened 2017-08-31
Fixed using wrong parameter for dictionaryForPlace

kiroskirin:


nickpatrick:
Good catch! I will merge and push a new release later today. Thank you.

PR #21, opened 2018-06-03
Add alternate places support to events

### LINKED ISSUE #20: Allow access to "alternate places" in react native
So far, radar looks really promising! However, it doesn't seem as if the react native bridge supports the ability to see alternate places. 

In Radar's documentation for iOS and Android you see the following code: `Radar.acceptEventId(event._id, verifiedPlaceId: event.alternatePlaces[0]._id)`. This indicates that the native SDK event includes an alternatePlaces array. However, in react-native, when listening for events, I see the following data for the event:
entered_place

However, when I view this event online (I verified that the id is the same), I see that it does show alternate places in the portal:
radar online

As you can see in the screenshot, these aren't passed to my react native application's event handler. (See below for code). log is a function I made to display the event data to me as I test it out.

```js
    Radar.on('events', (result: RadarOnEvent) => {
      // do something with result.events, result.user
      result.events.forEach((event: RadarEvent) => {
        log('EVENT: ' + event.type, event)
      })
    })
```
Interestingly, in the `index.js` for the `react-native-radar` module, there is [code which allows you to accept or reject an event ID. ](https://github.com/radarlabs/react-native-radar/blob/59e8ecd3c1e8e8e870decbaad8285d57da0bcaa0/js/index.js#L45-L51)

I'm not sure if it's a bug or if I haven't configured it properly, but since all the events seem to be firing correctly, I figured I would ask about it here.

ntomallen:
Seeing that the native modules were returning `event.alternatePlaces` already, it was really straightforward to add methods which add the `alternatePlaces` array to the event being returned to the react-native event emitter. I have tested this for both iOS and Android and seems to work properly. This resolves issue #20.

### DIFF
--- a/android/src/main/java/com/onradar/react/RNRadarUtils.java
+++ b/android/src/main/java/com/onradar/react/RNRadarUtils.java
@@ -276,11 +276,11 @@ class RNRadarUtils {
         return map;
     }
 
-    private static arrayForAlternatePlaces(RadarPlace[] places) {
+    private static WritableArray arrayForAlternatePlaces(RadarPlace[] places) {
         if (places == null) {
             return null;
         }
-    
+
         WritableArray arr = Arguments.createArray();
         for (RadarPlace place : places) {
             WritableMap map = RNRadarUtils.mapForPlace(place);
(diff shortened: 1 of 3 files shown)


ntomallen:
fix

Issue #23, opened 2018-06-13
Build fails because of typo


ntomallen:
I apologize for introducing the typo in the first place. It looks like there is still one there on line 277 that didn't make it into the fix in [that commit. ](https://github.com/radarlabs/react-native-radar/commit/320e4fc701bfddf9390edf5bd53bea9582ea864b)

[Here is the bad line. 
](https://github.com/radarlabs/react-native-radar/blob/36daa4df8fb345deb9fb03e0156911caa9e78c26/ios/RNRadarUtils.m#L277)

This causes the build to fail completely after updating on a new repository.

coolbrow:
Fixed in `1.0.17`

PR #24, opened 2018-06-13
Fix the typo preventing it from building

ntomallen:
See #23.


coolbrow:
Whoops, this was my mistake, thanks for the fix!

PR #37, opened 2018-08-14
Expose geofence metadata

nickpatrick:


android/src/main/java/io/radar/react/RNRadarUtils.java
```diff
@@ -98,6 +100,26 @@ static int numberForUserInsightsLocationConfidence(RadarUserInsightsLocation.Rad
         return Radar.RadarPlacesProvider.NONE;
     }
 
+    static WritableMap mapForJSONObject(JSONObject obj) {
+        WritableMap map = Arguments.createMap();
+        Iterator<String> keys = obj.keys();
+        while(keys.hasNext()) {
+            String key = keys.next();
+            Object val = obj.get(key);
+            if (obj instanceof String) {
```
coolbrow: All these conditionals should be checking against `val` not `obj`

android/src/main/java/io/radar/react/RNRadarUtils.java
```diff
@@ -98,6 +100,26 @@ static int numberForUserInsightsLocationConfidence(RadarUserInsightsLocation.Rad
...
+                String valStr = (String)val;
+                map.putString(key, valStr);
+            } else if (obj instanceof Number) {
+                Number valNum = (Number)val;
+                map.putInt(key, valNum.intValue());
+            } else if (obj instanceof Boolean) {
+                Boolean valBool = (Boolean)obj;
```
coolbrow: and `val` here too

android/src/main/java/io/radar/react/RNRadarUtils.java
```diff
@@ -98,6 +100,26 @@ static int numberForUserInsightsLocationConfidence(RadarUserInsightsLocation.Rad
...
+        Iterator<String> keys = obj.keys();
+        while(keys.hasNext()) {
+            String key = keys.next();
+            Object val = obj.get(key);
+            if (val instanceof String) {
+                String valStr = (String)val;
+                map.putString(key, valStr);
+            } else if (val instanceof Number) {
+                Number valNum = (Number)val;
+                map.putInt(key, valNum.intValue());
```
coolbrow: Do we only allow `int` values for numbers? `Double` might be better here

android/src/main/java/io/radar/react/RNRadarUtils.java
```diff
@@ -98,6 +100,26 @@ static int numberForUserInsightsLocationConfidence(RadarUserInsightsLocation.Rad
...
+            } else if (val instanceof Boolean) {
+                Boolean valBool = (Boolean)val;
+                map.putBoolean(valBool.booleanValue());
```
coolbrow: Shouldn't this have a key?

android/src/main/java/io/radar/react/RNRadarUtils.java
nickpatrick: We do allow doubles, but I think ints are more common. We could use double here since it's more flexible and tell folks to convert to int if more appropriate.

PR #39, opened 2018-10-30
Issue 18 - [iOS] Return "GRANTED_ALWAYS" when status is Always Allow and  [Android] "DENIED_FOREVER" when user selects 'never ask again' 

RodrigoCS:
This was a simple change I'm proposing "GRANTED_ALWAYS" as a string constant when the status is kCLAuthorizationStatusAuthorizedAlways.

I'm guessing it was previously just "GRANTED" to keep consistency between both platforms but since Android doesn't really need it and it doesn't even use UNKNOWN, to begin with I don't think this should

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,8 @@ Radar.getPermissionsStatus().then((status) => {
 - `GRANTED`
 - `GRANTED_ALWAYS` (iOS)
 - `DENIED`
-- `UNKNOWN`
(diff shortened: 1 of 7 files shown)


RodrigoCS:
Added a new permission when the user selects the never ask again checkbox

coolbrow:
Sorry for the radio silence on this one. We've talked about this API internally and have some plans for future improvements here.

jwebcat:
this is a feature I would like to see as well, also a callback from requestPermissions(background)

Any progress on this?

PR #47, opened 2018-11-28
Expose the new RadarTrackingOptions from SDK 2.1

coolbrow:
This allows users to apply the new options like so:
```js
Radar.startTracking({
  sync: 'all', // or 'possibleStateChanges'
  offline: 'replayStopped' // or 'replayOff'
})
```

### DIFF
--- a/android/src/main/java/io/radar/react/RNRadarModule.java
+++ b/android/src/main/java/io/radar/react/RNRadarModule.java
@@ -14,9 +14,6 @@ import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import io.radar.sdk.Radar;
-import io.radar.sdk.Radar.RadarTrackingOffline;
-import io.radar.sdk.Radar.RadarTrackingSync;
-import io.radar.sdk.RadarTrackingOptions;
 import io.radar.sdk.Radar.RadarCallback;
 import io.radar.sdk.model.RadarEvent;
 import io.radar.sdk.model.RadarUser;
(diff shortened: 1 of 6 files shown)


js/index.js
```diff
@@ -26,9 +26,9 @@ const requestPermissions = (background) => {
   NativeModules.RNRadar.requestPermissions(background);
 };
 
-const startTracking = () => {
-  NativeModules.RNRadar.startTracking();
-};
+const startTracking = (options) => {
+  NativeModules.RNRadar.startTracking(options);
+}
```
nickpatrick: I would keep the semicolon here for consistency. We may want to add linting at some point.

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -60,8 +63,31 @@ public void requestPermissions(boolean background) {
     }
 
     @ReactMethod
-    public void startTracking() {
-        Radar.startTracking();
+    public void startTracking(ReadableMap optionsMap) {
+        RadarTrackingOptions.Builder options = new RadarTrackingOptions.Builder();
```
nickpatrick: Looks good. May want to consider moving this logic to `RNRadarUtils`, same pattern as setting places provider.

ios/RNRadar.m
```diff
@@ -80,8 +80,28 @@ - (void)didFailWithStatus:(RadarStatus)status {
     }
 }
 
-RCT_EXPORT_METHOD(startTracking) {
-    [Radar startTracking];
+RCT_EXPORT_METHOD(startTracking:(NSDictionary *)optionsDict) {
+    RadarTrackingOptions *options = [RadarTrackingOptions new];
```
nickpatrick: May want to consider moving this logic to `RNRadarUtils`.

PR #48, opened 2018-12-06
Add `requiresMainQueueSetup` to RNRadar

sairus2k:
Fix warning in iOS:
>Module RNRadar requires main queue setup since it overrides `init` but doesn't implement `requiresMainQueueSetup`. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.


coolbrow:
Are you on the latest version? This method already exists here - https://github.com/radarlabs/react-native-radar/blob/master/ios/RNRadar.m#L25

PR #86, opened 2019-10-08
iOS Efficiency Mode 

brettguenther:
- Bump required Radar native iOS SDK version to 2.1.6
- Handle priority key in tracking options dictionary


ios/RNRadarTests/RNRadarUtilsTests.m
```diff
@@ -73,12 +73,13 @@ - (void)testPlacesProviderForString {
 }
 
 - (void)testOptionsForDictionary {
-    NSDictionary *optionsDict = @{@"sync": @"all", @"offline": @"replayOff", @"invalid": @"ingore me"};
+    NSDictionary *optionsDict = @{@"sync": @"all", @"offline": @"replayOff", @"invalid": @"ingore me", @"priority":@"responsiveness"};
```
nickpatrick: @brettguenther Can you add a space `@"priority": @"responsiveness"`? That will also trigger another build which should cause checks to pass now.

PR #90, opened 2020-02-24
v3 support

nickpatrick:
- Upgrade to Radar `3.0.0` and React Native `0.60`
- Remove model-specific serialization helpers from `RNRadarUtils`, use new native helpers instead
- Support background location permission requests and constants
- Add example app, simplify tests

- Most of the magic happens in:
  - `js/index.js`
  - `ios/RNRadar.m`
  - `android/src/main/java/io/radar/react/RNRadarModule.java`


README.md
```diff
@@ -2,232 +2,16 @@
 
 [![npm](https://img.shields.io/npm/v/react-native-radar.svg)](https://www.npmjs.com/package/react-native-radar)
 
-[Radar](https://radar.io) is the location platform for mobile apps.
+[Radar](https://radar.io) is location data infrastructure. You can use Radar SDKs and APIs to add location context to your apps with just a few lines of code.
 
-## Installation
```
nickpatrick: I'm going to move the docs into Contentful

nickpatrick:
@kochis @jsani-radar Ready for review, happy to walk you through it if helpful

.circleci/config.yml
```diff
@@ -6,112 +6,44 @@ jobs:
...
-
-      - store_test_results:
-          path: test-results
-
-      - store_artifacts:
-          path: test-results
-
+          path: ~/react-native-radar
+      - run: yarn install
+      - run: sudo npm install -g react-native
```
jsani-radar: seems odd at first glance to do a `yarn install` and then a `npm install`, are both package managers used here?

.circleci/config.yml
```diff
@@ -6,112 +6,44 @@ jobs:
...
     steps:
-      - checkout
-
-      - restore_cache:
-          name: Restore node modules cache
-          key: lock-v1-{{ checksum "package-lock.json" }}-{{ arch }}
-
-      - run: npm install
- 
-      - save_cache:
```
jsani-radar: do we not want to cache installed modules here? I guess we can always add this back later

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -42,59 +56,99 @@ public void setDescription(String description) {
...
-    }
-
-    @ReactMethod
-    public void setPlacesProvider(String providerStr) {
-        Radar.setPlacesProvider(RNRadarUtils.placesProviderForString(providerStr));
+        Radar.setMetadata(RNRadarUtils.jsonForMap(metadataMap));
     }
 
     @ReactMethod
     public void getPermissionsStatus(Promise promise) {
```
jsani-radar: seems odd to pass a `Promise` as a param to a function rather than have it as a return type -- I guess if that's the RN way 🤷‍♂

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -42,59 +56,99 @@ public void setDescription(String description) {
...
 
     @ReactMethod
     public void trackOnce(final Promise promise) {
-        Radar.trackOnce(new RadarCallback() {
+        Radar.trackOnce(new Radar.RadarTrackCallback() {
             @Override
-            public void onComplete(@NonNull Radar.RadarStatus status, Location location, RadarEvent[] events, RadarUser user) {
-                if (promise == null)
+            public void onComplete(@NonNull Radar.RadarStatus status, @Nullable Location location, @Nullable RadarEvent[] events, @Nullable RadarUser user) {
+                if (promise == null) {
```
jsani-radar: wonder if this is the desirable default behavior if a `Promise` is not passed here - we can still execute the request / operation without resolving / rejecting back to the caller. I believe the Web SDK does it this way

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -142,4 +236,449 @@ public void rejectEvent(String eventId) {
...
+            public void onComplete(@NonNull Radar.RadarStatus status, @Nullable Location location, @Nullable RadarContext context) {
+                if (status == Radar.RadarStatus.SUCCESS) {
+                    try {
+                        WritableMap map = Arguments.createMap();
+                        map.putString("status", status.toString());
+                        if (location != null) {
+                            map.putMap("location", RNRadarUtils.mapForJson(Radar.jsonForLocation(location)));
+                        }
+                        if (context != null) {
+                            map.putMap("user", RNRadarUtils.mapForJson(context.toJson()));
```
jsani-radar: nit: the top-level key we use in the API responses is `context` - https://github.com/radarlabs/server/blob/master/app.ts#L647

might be confusing if we key by `user` if this is meant to be a stateless operation

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -142,4 +236,449 @@ public void rejectEvent(String eventId) {
...
+            public void onComplete(@NonNull Radar.RadarStatus status, @Nullable Location location, @Nullable RadarContext context) {
+                if (status == Radar.RadarStatus.SUCCESS) {
+                    try {
+                        WritableMap map = Arguments.createMap();
+                        map.putString("status", status.toString());
+                        if (location != null) {
+                            map.putMap("location", RNRadarUtils.mapForJson(Radar.jsonForLocation(location)));
+                        }
+                        if (context != null) {
+                            map.putMap("context", RNRadarUtils.mapForJson(context.toJson()));
```
jsani-radar: ditto: consistency with keys above

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -142,4 +236,449 @@ public void rejectEvent(String eventId) {
...
+        }
+    }
+
+    @ReactMethod
+    public void autocomplete(ReadableMap optionsMap, final Promise promise) {
+        if (promise == null) {
+            return;
+        }
+
+
```
jsani-radar: nit: extra `\n`

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -142,4 +236,449 @@ public void rejectEvent(String eventId) {
...
+            @Override
+            public void onComplete(@NonNull Radar.RadarStatus status, @Nullable RadarAddress[] addresses) {
+                if (status == Radar.RadarStatus.SUCCESS) {
+                    try {
+                        WritableMap map = Arguments.createMap();
+                        map.putString("status", status.toString());
+                        if (addresses != null) {
+                            map.putArray("addresses", RNRadarUtils.arrayForJson(RadarAddress.toJson(addresses)));
+                        }
+                        promise.resolve(map);
```
jsani-radar: do we want to resolve if `addresses == null`? that seems like an error on our side?

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -142,4 +236,449 @@ public void rejectEvent(String eventId) {
...
```
jsani-radar: ditto

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -142,4 +236,449 @@ public void rejectEvent(String eventId) {
...
+        if (promise == null) {
+            return;
+        }
+
+        Radar.ipGeocode(new Radar.RadarIpGeocodeCallback() {
+            @Override
+            public void onComplete(@NonNull Radar.RadarStatus status, @Nullable RadarAddress address) {
+                if (status == Radar.RadarStatus.SUCCESS) {
+                    try {
+                        WritableMap map = Arguments.createMap();
```
jsani-radar: seems like we repeat this pattern quite often, not necessary for this version, but we can probably abstract the deserialization of a JSON key / checking nullability - ended up coming out pretty clean in the Python SDK https://github.com/radarlabs/radar-python/blob/5b9a61bda3e405565eb16d76b120cc27f7a2a7b3/radar/endpoints.py#L80-L84

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -142,4 +236,449 @@ public void rejectEvent(String eventId) {
...
+
+        ReadableMap originMap = optionsMap.getMap("origin");
+        Location origin = null;
+        if (originMap != null) {
+            double latitude = originMap.getDouble("latitude");
+            double longitude = originMap.getDouble("longitude");
+            origin = new Location("RNRadarModule");
+            origin.setLatitude(latitude);
+            origin.setLongitude(longitude);
+        }
```
jsani-radar: do we still want to execute the request if `origin == null`? seems like the code will still fall through here..

android/src/main/java/io/radar/react/RNRadarReceiver.java
```diff
@@ -78,25 +78,47 @@ public void onLocationUpdated(@NonNull Context context, @NonNull Location locati
...
+            reactNativeHost = reactApplication.getReactNativeHost();
+
+            WritableMap map = Arguments.createMap();
+            map.putMap("location", RNRadarUtils.mapForJson(Radar.jsonForLocation(location)));
+            map.putBoolean("stopped", stopped);
+            map.putString("source", source.toString());
+
+            sendEvent("clientLocation", map);
+        } catch (Exception e) {
+
```
jsani-radar: empty `catch` block?

ios/RNRadar.m
```diff
@@ -31,7 +29,7 @@ - (dispatch_queue_t)methodQueue {
 }
 
 - (NSArray<NSString *> *)supportedEvents {
-    return @[@"events", @"location", @"error"];
+    return @[@"events", @"location", @"clientLocation", @"error"];
```
jsani-radar: what's the difference between `location` and `clientLocation` here?

ios/RNRadar.m
```diff
@@ -44,22 +42,32 @@ - (void)stopObserving {
...
 - (void)didUpdateLocation:(CLLocation *)location user:(RadarUser *)user {
     if (hasListeners) {
-        [self sendEventWithName:@"location" body:@{@"location": [RNRadarUtils dictionaryForLocation:location], @"user": [RNRadarUtils dictionaryForUser:user]}];
+        [self sendEventWithName:@"location" body:@{@"location": [Radar dictionaryForLocation:location], @"user": [user dictionaryValue]}];
+    }
+}
+
+- (void)didUpdateClientLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarLocationSource)source {
+    if (hasListeners) {
+        [self sendEventWithName:@"clientLocation" body:@{@"location": [Radar dictionaryForLocation:location], @"stopped": @(stopped), @"source": [Radar stringForSource:source]}];
```
jsani-radar: nit: JSON body might be easier to read if we tab it down

ios/RNRadar.m
```diff
@@ -44,22 +42,32 @@ - (void)stopObserving {
...
 
 - (void)didFailWithStatus:(RadarStatus)status {
     if (hasListeners) {
-        [self sendEventWithName:@"error" body:[RNRadarUtils stringForStatus:status]];
+        [self sendEventWithName:@"error" body:[Radar stringForStatus:status]];
     }
 }
 
+- (void)didLogMessage:(NSString *)message {
+
```
jsani-radar: empty?

ios/RNRadar.m
```diff
@@ -88,79 +111,539 @@ - (void)didFailWithStatus:(RadarStatus)status {
...
+        if (reject) {
+            reject([Radar stringForStatus:RadarStatusErrorBadRequest], [Radar stringForStatus:RadarStatusErrorBadRequest], nil);
+        }
+
+        return;
+    }
+
+    NSDictionary *nearDict = optionsDict[@"near"];
+    CLLocation *near;
+    if (nearDict != nil && [nearDict isKindOfClass:[NSDictionary class]]) {
```
jsani-radar: looks like `near` is required on the server - https://github.com/radarlabs/server/blob/master/app.ts#L565

do we want to do any upstream validation or let it 400 over the wire?

ios/RNRadar.m
```diff
@@ -88,79 +111,539 @@ - (void)didFailWithStatus:(RadarStatus)status {
...
+    double latitude = [RCTConvert double:nearDict[@"latitude"]];
+    double longitude = [RCTConvert double:nearDict[@"longitude"]];
+    NSDate *timestamp = [NSDate new];
+    CLLocation *near = [[CLLocation alloc] initWithCoordinate:CLLocationCoordinate2DMake(latitude, longitude) altitude:-1 horizontalAccuracy:5 verticalAccuracy:-1 timestamp:timestamp];
+    NSNumber *limitNumber = optionsDict[@"limit"];
+    int limit;
+    if (limitNumber != nil && [limitNumber isKindOfClass:[NSNumber class]]) {
+        limit = [limitNumber intValue];
+    } else {
+        limit = 10;
```
jsani-radar: we shim 100 as the limit on the server if none is provided

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: Yeah, this is standard for RN modules. Docs: https://reactnative.dev/docs/native-modules-android

.circleci/config.yml
nickpatrick: I was running into weirdness with caching, will see if I can add it back

.circleci/config.yml
nickpatrick: `yarn` used to install dependencies, `npm` used to install `react-native` command line tools for builds. I'll see if I can just install it with `yarn` for consistency

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: In this case, we are calling `trackOnce` even if promise is `null`. We're just skipping resolving/rejecting the promise. It's valid to call `trackOnce` to force a location update even if you're not handling the response.

For the other APIs, there's no point in making an API request if you're not going to handle the response, so we just short circuit. (See `getContext()`, for example.)

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: Good catch, this should be `context`. Fixing.

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: If status is success, should never be the case. Mainly just guarding for any downstream crashes

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: If `origin` is `null`, we call the variant that gets the device's currently location and uses it for `origin`. See https://github.com/radarlabs/react-native-radar/pull/90/files/d3bf9d980e5a8d9ea7923b1659b424025fc6140a#diff-d5f0c15b95a2ea3236e058bc4a2d47b8R678.

ios/RNRadar.m
nickpatrick: `location` is for the `didUpdateLocation` callback (synced to server), `clientLocation` is for the `didUpdateClientLocation` callback (updated on client but not necessarily synced to server)

ios/RNRadar.m
nickpatrick: Will do

ios/RNRadar.m
nickpatrick: Will add for completeness

ios/RNRadar.m
nickpatrick: Seems high. We may want to change default to 10. Will Clubhouse

ios/RNRadar.m
nickpatrick: Look a little bit lower. If `near` is provided, we call `searchPlacesNear:` (provided location), otherwise we call `searchPlacesWithRadius:` (get current device location)

.circleci/config.yml
nickpatrick: Challenge with caching here is that the example app pulls the module code from the parent directory instead of npm. We want it to pull a fresh version every time, caching messed with that. Could probably get smarter about this but this doesn't take terribly long to run, so I think we're fine for now.

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: Yeah, we can refactor this in a second pass if needed

android/src/main/java/io/radar/react/RNRadarReceiver.java
nickpatrick: Added logging here

nickpatrick:
@jsani-radar Worth another look when you're ready. Any inconsistencies in params format for new APIs between web SDK and RN module? Ideally those are consistent.

`example/App.js` (https://github.com/radarlabs/react-native-radar/blob/785e135a5e7de02a91506237b5e132ae9bfea069/example/App.js) and `test/index.test.js` (https://github.com/radarlabs/react-native-radar/blob/785e135a5e7de02a91506237b5e132ae9bfea069/test/index.test.js) are good places to quickly glance at sample code.

cc @kochis as well

.circleci/config.yml
jsani-radar: I see, we can probably check the ts or checksum on the parent directory to see if a fresh pull is needed - agreed not a big deal for this iteration

.circleci/config.yml
```diff
@@ -6,113 +6,44 @@ jobs:
...
-      - store_test_results:
-          path: test-results
-
-      - store_artifacts:
-          path: test-results
-
+      - checkout:
+          path: ~/react-native-radar
+      - run: yarn install
+      - run: sudo yarn global add react-native
```
jsani-radar: 🎉

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -42,59 +56,99 @@ public void setDescription(String description) {
...
+            } else {
+                promise.resolve("DENIED");
+            }
+        } else {
+            promise.resolve(foreground ? "GRANTED_BACKGROUND" : "DENIED");
+        }
     }
 
     @ReactMethod
     public void requestPermissions(boolean background) {
```
jsani-radar: do we need the `background` param here?

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -109,29 +163,69 @@ public void updateLocation(ReadableMap locationMap, final Promise promise) {
...
+        Radar.startTracking(RadarTrackingOptions.CONTINUOUS);
+    }
+
+    @ReactMethod
+    public void startTrackingCustom(ReadableMap optionsMap) {
+        try {
+            JSONObject optionsObj = RNRadarUtils.jsonForMap(optionsMap);
+            RadarTrackingOptions options = RadarTrackingOptions.fromJson(optionsObj);
+            Radar.startTracking(options);
+        } catch (JSONException e) {
```
jsani-radar: is it possible for `json` serialization of an `optionsMap` to error out?

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: Good catch, yes, we need to check this before requesting background permissions. Will add.

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: Yeah, the `optionsMap` methods can throw `JSONException`. (I'm not sure why React Native uses these instead of `JSONObject`s in the first place...)

jsani-radar: yeah, definitely odd, let's bubble up this exception to the caller then?

PR #114, opened 2020-06-24
Add geofence search metadata, remove transit, bump versions

nickpatrick:


nickpatrick:
@tjulien Build should pass once I ship native SDKs

PR #119, opened 2020-07-16
Trip and mock tracking

nickpatrick:


android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -242,6 +304,22 @@ public void rejectEvent(String eventId) {
...
     }
 
+    @ReactMethod
+    public void startTrip(ReadableMap optionsMap) {
+        try {
+            JSONObject optionsObj = RNRadarUtils.jsonForMap(optionsMap);
+            RadarTripOptions options = RadarTripOptions.fromJson(optionsObj);
+            Radar.startTrip(options);
+        } catch (JSONException e) {
+
```
jsani-radar, line 314: report the error back to the caller here?

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -227,6 +228,67 @@ public void startTrackingCustom(ReadableMap optionsMap) {
...
+        ReadableMap originMap = optionsMap.getMap("origin");
+        double originLatitude = originMap.getDouble("latitude");
+        double originLongitude = originMap.getDouble("longitude");
+        Location origin = new Location("RNRadarModule");
+        origin.setLatitude(originLatitude);
+        origin.setLongitude(originLongitude);
+        ReadableMap destinationMap = optionsMap.getMap("destination");
+        double destinationLatitude = destinationMap.getDouble("latitude");
+        double destinationLongitude = destinationMap.getDouble("longitude");
```
jsani-radar, line 245: would be nice to have utils for these at some point like the collection additions on iOS, not necessary now

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -242,6 +304,22 @@ public void rejectEvent(String eventId) {
...
+            RadarTripOptions options = RadarTripOptions.fromJson(optionsObj);
+            Radar.startTrip(options);
+        } catch (JSONException e) {
+
+        }
+    }
+
+    @ReactMethod
+    public void stopTrip() {
+        Radar.stopTrip();
```
jsani-radar, line 320: wrap in a `try / catch` as well?

nickpatrick: Don't need to wrap in a `try/catch` because we're not manipulating a `JSONObject` here

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: `startTrip()` doesn't accept a callback, but let me add some error logging for exceptions like we have in the Cordova lib.

PR #127, opened 2021-02-12
Expose set log level

kochis:
iOS and Android SDK's support setting the SDK log level: https://radarlabs.github.io/radar-sdk-ios/Classes/Radar.html#/c:objc(cs)Radar(cm)setLogLevel:

This exposes that hook in React Native via `Radar.setLogLevel`

### DIFF
--- a/android/src/main/java/io/radar/react/RNRadarModule.java
+++ b/android/src/main/java/io/radar/react/RNRadarModule.java
@@ -19,6 +19,7 @@ import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import io.radar.sdk.Radar;
+import io.radar.sdk.Radar.RadarLogLevel
 import io.radar.sdk.RadarTrackingOptions;
 import io.radar.sdk.RadarTripOptions;
 import io.radar.sdk.model.RadarAddress;
--- a/android/src/main/java/io/radar/react/RNRadarModule.java
+++ b/android/src/main/java/io/radar/react/RNRadarModule.java
@@ -770,4 +770,8 @@ public class RNRadarModule extends ReactContextBaseJavaModule {
         }
     }
 
+    @ReactMethod
+    public void setLogLevel(int level) {
+      Radar.setLogLevel(level);
+    }
 }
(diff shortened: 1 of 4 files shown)


android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -19,6 +19,7 @@
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import io.radar.sdk.Radar;
+import io.radar.sdk.Radar.RadarLogLevel
```
nickpatrick, line 22: @kochis May need a `;`

ios/RNRadar.m
```diff
@@ -81,6 +81,10 @@ - (void)didLogMessage:(NSString *)message {
     }
 }
 
+RCT_EXPORT_METHOD(setLogLevel:(NSInteger *)level) {
+    [Radar setLogLevel:level];
```
nickpatrick, line 85: @kochis May want to cast this to `(RadarLogLevel)` (although may not strictly be necessary)

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -63,6 +64,11 @@ public void setMetadata(ReadableMap metadataMap) throws JSONException {
         Radar.setMetadata(RNRadarUtils.jsonForMap(metadataMap));
     }
 
+    @ReactMethod
+    public void setLogLevel(int level) {
+      Radar.setLogLevel(Radar.RadarLogLevel.fromInt(level));
```
jsani-radar, line 69: hmm does this `fromInt` mapping fn exist? seems like that's what the build is complaining about?

nickpatrick: @jsani-radar Yep, it's there in Kotlin, just not exposed in Java. Needs a `@JvmStatic` annotation which we're adding in v3.1. That said, wonder if we should support passing a string instead of an int and use an if statement to determine the correct enum value? @brettguenther did this in our new Flutter plugin and it may be more readable/usable. In other words, support `Radar.setLogLevel('info')` instead of `Radar.setLogLevel(3)`

nickpatrick: Now that we've shipped the Flutter SDK, let's just support passing a string instead (see Flutter impl: https://github.com/radarlabs/flutter-radar/blob/master/ios/Classes/RadarFlutterPlugin.m#L136). We can do the same across all of the JS libs including Cordova and Capacitor. Bypasses this issue and is a better developer experience anyway I think.

nickpatrick:
Closing in favor of #139

PR #146, opened 2021-08-09
Add NOT_DETERMINED permissions state

nickpatrick:
- Adds `NOT_DETERMINED` permissions state on iOS and Android
- Bumps versions
- Adds back `Podfile.lock` and updates `metro` config for example app

### DIFF
--- a/android/src/main/java/io/radar/react/RNRadarModule.java
+++ b/android/src/main/java/io/radar/react/RNRadarModule.java
@@ -96,7 +96,7 @@ public class RNRadarModule extends ReactContextBaseJavaModule {
         }
 
         boolean foreground = ActivityCompat.checkSelfPermission(activity, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
-        boolean background = true;
+        boolean background = foreground;
         boolean denied = ActivityCompat.shouldShowRequestPermissionRationale(activity, Manifest.permission.ACCESS_FINE_LOCATION);
         
         if (Build.VERSION.SDK_INT >= 29) {
--- a/android/src/main/java/io/radar/react/RNRadarModule.java
+++ b/android/src/main/java/io/radar/react/RNRadarModule.java
@@ -87,11 +87,14 @@ public class RNRadarModule extends ReactContextBaseJavaModule {
             return;
         }
 
+        Activity activity = getCurrentActivity();
         boolean foreground = ActivityCompat.checkSelfPermission(getReactApplicationContext(), Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
         if (Build.VERSION.SDK_INT >= 29) {
             if (foreground) {
                 boolean background = ActivityCompat.checkSelfPermission(getReactApplicationContext(), Manifest.permission.ACCESS_BACKGROUND_LOCATION) == PackageManager.PERMISSION_GRANTED;
                 promise.resolve(background ? "GRANTED_BACKGROUND" : "GRANTED_FOREGROUND");
+            } else if (ActivityCompat.shouldShowRequestPermissionRationale(activity, Manifest.permission.ACCESS_FINE_LOCATION)) {
+                promise.resolve("NOT_DETERMINED");
             } else {
                 promise.resolve("DENIED");
             }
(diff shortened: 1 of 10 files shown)


example/metro.config.js
```diff
@@ -1,5 +1,17 @@
+const blacklist = require('metro-config/src/defaults/blacklist');
```
nickpatrick, line 1: Ran into this: https://github.com/facebook/react-native/issues/24065#issuecomment-537489786

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -87,16 +87,24 @@ public void getPermissionsStatus(Promise promise) {
             return;
         }
 
-        boolean foreground = ActivityCompat.checkSelfPermission(getReactApplicationContext(), Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
+        Activity activity = getCurrentActivity();
+
+        boolean foreground = ActivityCompat.checkSelfPermission(activity, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
+        boolean background = true;
+        boolean denied = ActivityCompat.shouldShowRequestPermissionRationale(activity, Manifest.permission.ACCESS_FINE_LOCATION);
```
nickpatrick, line 94: `shouldShowRequestPermissionRationale()` returns `true` if a permission prompt has been previously shown and permissions were denied: https://developer.android.com/training/permissions/requesting#explain

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -87,16 +87,24 @@ public void getPermissionsStatus(Promise promise) {
             return;
         }
 
-        boolean foreground = ActivityCompat.checkSelfPermission(getReactApplicationContext(), Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
+        Activity activity = getCurrentActivity();
+
+        boolean foreground = ActivityCompat.checkSelfPermission(activity, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
+        boolean background = true;
```
nickpatrick, line 93: There was no separate background permission before API level 29, this flag is updated below

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -87,16 +87,24 @@ public void getPermissionsStatus(Promise promise) {
             return;
         }
 
-        boolean foreground = ActivityCompat.checkSelfPermission(getReactApplicationContext(), Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
+        Activity activity = getCurrentActivity();
```
tjulien, line 90: we should probably do a `null` check on `activity`

react-native-radar.podspec
```diff
@@ -15,5 +15,5 @@ Pod::Spec.new do |s|
   s.platform = :ios, "10.0"
 
   s.dependency "React"
-  s.dependency "RadarSDK", "3.1.7"
+  s.dependency "RadarSDK", "3.1.10"
```
tjulien, line 18: not `3.1.14`, like in `android/build.gradle`?

example/metro.config.js
tjulien: let's add this as a comment here

react-native-radar.podspec
nickpatrick: Nope, iOS is only on `3.1.10`

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: Will do

example/metro.config.js
nickpatrick: Done

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: Done

android/src/main/java/io/radar/react/RNRadarModule.java
tjulien: should this be `boolean background = foreground;`?  Otherwise on VERSION < 29 it's always = `true` and we'll always `promise.resolve("GRANTED_BACKGROUND");`

nickpatrick: That's the intended behavior since there was no separate background permission before API level 29

tjulien: if you denied foreground perms, you would still have background perms?

nickpatrick: Ohhh I see what you're saying now. Will update

nickpatrick: Good catch, done

tjulien (approved):

PR #150, opened 2021-08-15
Permission callbacks

nickpatrick:
Support promises for permission requests like

```
Radar.requestPermissions(false).then((result) => {
  // called after permission prompt dismissed, or immediately if prompt not shown
});
  ```

### DIFF
--- a/android/src/main/java/io/radar/react/RNRadarModule.java
+++ b/android/src/main/java/io/radar/react/RNRadarModule.java
@@ -121,7 +121,7 @@ public class RNRadarModule extends ReactContextBaseJavaModule implements Permiss
     @Override
     public boolean onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
         if (requestCode == PERMISSIONS_REQUEST_CODE && mPermissionsRequestPromise != null) {
-            getPermissionStatus(mPermissionsRequestPromise);
+            getPermissionsStatus(mPermissionsRequestPromise);
             mPermissionsRequestPromise = null;
         }
         return true;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by tjulien** [comment]
https://app.clubhouse.io/radarlabs/story/6965/permission-promises-for-react-native

File: android/src/main/java/io/radar/react/RNRadarModule.java [hunk 1]
```diff
@@ -36,9 +38,11 @@
 
 import java.util.EnumSet;
 
-public class RNRadarModule extends ReactContextBaseJavaModule {
+public class RNRadarModule extends ReactContextBaseJavaModule implements PermissionListener {
 
     private static final String TAG = "RNRadarModule";
+    private static final int PERMISSIONS_REQUEST_CODE = 20160525;
```
tjulien, line 44: where does this constant come from?

nickpatrick: Just a random int that we use for `requestCode`. `2016-05-25` is the date Radar was incorporated, so we use it all over the place (e.g., https://github.com/radarlabs/radar-sdk-android/blob/abbdda744e0de4d44c71f7f1612c2d0cb5999d05/sdk/src/main/java/io/radar/sdk/RadarJobScheduler.kt#L27)

tjulien: gotcha, yeah let's just add a comment above it, something like:

```
// random request code (Radar's birthday!)
```

So people know where it comes from

tjulien (approved):

Issue #35, opened 2018-08-09
RnRadarReceiver Crash on Android (In Background?)


ntomallen:
We've deployed an update to our internal testers with the new Radar package. We have crashlytics enabled, so I get detailed crash reports for apps in release mode. I just got two reports for a crash with **RNRadarReceiver.java**. The error is: `Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.content.BroadcastReceiver$PendingResult.finish()' on a null object reference`

Here is the stack trace:

```
io.radar.react.RNRadarReceiver$1.onReactContextInitialized (RNRadarReceiver.java:38)
com.facebook.react.ReactInstanceManager$6.run (ReactInstanceManager.java:984)
android.os.Handler.handleCallback (Handler.java:751)
android.os.Handler.dispatchMessage (Handler.java:95)
android.os.Looper.loop (Looper.java:154)
android.app.ActivityThread.main (ActivityThread.java:6123)
java.lang.reflect.Method.invoke (Method.java)
com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:867)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:757)
```

The device is:

```
Brand: motorola
Model: Moto G (5) Plus
RAM free: 911.12 MB
Disk free: 22.79 GB
```

The device reported That the app was in the background when it happened. Any ideas?

ntomallen:
Yeah, Android testers are also reporting both

1. No background location
2. Crashes while in the background

ntomallen:
Looking through the source code for Broadcast receiver here (see line 365): https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/java/android/content/BroadcastReceiver.java

This is the only online resource I can find related to it: https://stackoverflow.com/questions/49501196/broadcastreceiver-goasync-returns-null-pendingresult

It looks as though `goAsync()` gets `mPendingResult` and returns it, and then sets `mPendingResult` to null. So this could happen if `goAsync()` was called multiple times there..

It looks as though `setPendingResult()` might need to be called or something.. I'm not sure, the docs don't elaborate very much on it

coolbrow:
Hey @ntomallen, thanks for looking into this!

You are correct that the issue seems to be `goAsync()` returning `null` and causing this crash. Currently investigating this issue and should be resolved shortly.

[...]

PR #158, opened 2021-11-11
Update to Radar Android SDK 3.2.2

phil-brown:
These changes update the Android side of things - adds the Radar SDK 3.2.2 and updates the Gradle and AGP versions. It also adds some static analysis checks and updates Android Permission checks. Finally, this PR also adds support for setting the publishable key in strings.xml instead of in code.


tjulien:
https://app.shortcut.com/radarlabs/story/7670

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -99,12 +123,15 @@ public void getPermissionsStatus(final Promise promise) {
...
-        boolean foreground = ActivityCompat.checkSelfPermission(activity, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
+        boolean foreground = ActivityCompat.checkSelfPermission(activity,
+                Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
         boolean background = foreground;
-        boolean denied = ActivityCompat.shouldShowRequestPermissionRationale(activity, Manifest.permission.ACCESS_FINE_LOCATION);
-        
-        if (Build.VERSION.SDK_INT >= 29) {
-            background = ActivityCompat.checkSelfPermission(activity, Manifest.permission.ACCESS_BACKGROUND_LOCATION) == PackageManager.PERMISSION_GRANTED;
+        boolean denied = ActivityCompat.shouldShowRequestPermissionRationale(activity,
+                Manifest.permission.ACCESS_FINE_LOCATION);
```
brettguenther, line 130: since we are now targeting 31, do we need to handle approximate location in android 12?

kochis:
> This changes the recommended configuration that we should document in react docs, such that the method RNRadar.initialize should be used instead of Radar.initialize.

Is this a breaking change? Or will previous methods of initialization still work?

phil-brown:
@kochis yes. If the apps use Kotlin, then it will be less noticeable, since initialization of Radar without a receiver will work, as it will default to null. For Java, they will be forced to specify the receiver. They can either use `new RNRadarReceiver()` here, or use the `RNRadar.initialize` method. Do we need a migration guide? https://github.com/radarlabs/radar-sdk-android/blob/master/MIGRATION.md#31x-to-32x

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -215,15 +252,18 @@ public void trackOnce(ReadableMap locationMap, final Promise promise) {
...
         Location location = new Location("RNRadarModule");
         location.setLatitude(latitude);
         location.setLongitude(longitude);
         location.setAccuracy(accuracy);
 
         Radar.trackOnce(location, new Radar.RadarTrackCallback() {
             @Override
-            public void onComplete(@NonNull Radar.RadarStatus status, @Nullable Location location, @Nullable RadarEvent[] events, @Nullable RadarUser user) {
+            public void onComplete(@NonNull Radar.RadarStatus status,
+                                   @Nullable Location location1,
```
kochis, line 264: Why the change to `location1`?

android/src/main/java/io/radar/react/RNRadarModule.java
phil-brown: @brettguenther I made some updates that align better with this comment, but it would be good to understand if there are larger needs around these checks (using https://github.com/radarlabs/radar-sdk-android/pull/106 as a reference point)

android/src/main/java/io/radar/react/RNRadarModule.java
phil-brown: This occurred during some auto-refactors. I had tried using lambdas instead of callbacks in this module, however because the example app targets a lower android build tools version/AGP, lambdas aren't supported when consumed, so I changed back. As a result, some of these names changed. I just changed this one back - please lmk if you find others.

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -99,12 +135,24 @@ public void getPermissionsStatus(final Promise promise) {
             return;
         }
 
-        boolean foreground = ActivityCompat.checkSelfPermission(activity, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
+        int gpsLocation = ActivityCompat.checkSelfPermission(activity, Manifest.permission.ACCESS_FINE_LOCATION);
```
tjulien, line 138: how confident are we that `FINE` is synonymous with gps and `COARSE` is synonymous with wifi?  I do see some content out there claiming this.

tjulien: wondering if it might be more straightforward to prefix the variables with `fine` and `coarse` instead

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -129,14 +177,19 @@ public boolean onRequestPermissionsResult(int requestCode, String[] permissions,
...
+        PermissionAwareActivity activity = (PermissionAwareActivity) getCurrentActivity();
         mPermissionsRequestPromise = promise;
         if (activity != null) {
-            if (Build.VERSION.SDK_INT >= 23) {
-                if (background && Build.VERSION.SDK_INT >= 29) {
-                    activity.requestPermissions(new String[] { Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION }, PERMISSIONS_REQUEST_CODE, this);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                if (background && Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                    activity.requestPermissions(new String[]{
+                            Manifest.permission.ACCESS_FINE_LOCATION,
```
tjulien, line 186: I wonder if we need to start passing in `COARSE` here as well. 🤔   I bet the OS will just show the standard prompt with both `FINE` and `COARSE` even if you don't ask for `COARSE`.  I bet we're fine with the code as it is.

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -31,23 +40,38 @@
...
 
     private static final String TAG = "RNRadarModule";
-    private static final int PERMISSIONS_REQUEST_CODE = 20160525; // random request code (Radar's birthday!)
+    // random request code (Radar's birthday!)
+    private static final int PERMISSIONS_REQUEST_CODE = 20160525;
     private Promise mPermissionsRequestPromise;
 
     public RNRadarModule(ReactApplicationContext reactContext) {
         super(reactContext);
+        String publishableKey = reactContext.getString(R.string.radar_publishableKey);
```
tjulien, line 63: Nice new pattern.  So I guess now there are 3 ways to initialize the sdk?

1) configure `radar_publishableKey` in `strings.xml`
2) Call `Radar.initialize()` manually
3) Use the new `RNRadar` helper

Wondering if we want to have 3 different ways to do it, or if that's too many choices.  Could be fine as long as we document it and write a good migration guide.  1) Seems like far and away the best way for most apps to do it. cc @brettguenther @nickpatrick who have more experience with customer set ups here.

android/src/main/java/io/radar/react/RNRadarReceiver.java
```diff
@@ -18,36 +19,35 @@
...
         ReactContext reactContext = reactInstanceManager.getCurrentReactContext();
         if (reactContext == null) {
-            if (result == null) {
-                result = goAsync();
-            }
-            pendingCount.incrementAndGet();
-            reactInstanceManager.addReactInstanceEventListener(new ReactInstanceManager.ReactInstanceEventListener() {
+            reactInstanceManager.addReactInstanceEventListener(
+                    new ReactInstanceManager.ReactInstanceEventListener() {
+
```
tjulien, line 44: Are there any potential gotchas from not using a `PendingResult` anymore?

Was reading these:
https://developer.android.com/reference/android/content/BroadcastReceiver.PendingResult

@brettguenther or @nickpatrick may know more of the back story here
https://stackoverflow.com/questions/22741202/how-to-use-goasync-for-broadcastreceiver

tjulien: Here's the original PR and Github issue where the `PendingResult` was added:

https://github.com/radarlabs/react-native-radar/pull/36
https://github.com/radarlabs/react-native-radar/issues/35

android/src/main/java/io/radar/react/RNRadarModule.java
phil-brown: Good catch. I will add it. This code is all based off the sdk implementation, which does not request coarse permissions at the moment. There is a "permission-groups" feature that we should look into for all Android-based SDKs.

android/src/main/java/io/radar/react/RNRadarReceiver.java
phil-brown: `goAsync()` is a method on a BroadcastReceiver to ensure that long operations are not killed by the OS. Since this class no longer extends BroadcastReceiver, there's no PendingIntent to use.

android/src/main/java/io/radar/react/RNRadarModule.java
brettguenther: ah nice - see Tim commented on what i was thinking as well with including ACCESS_COARSE_LOCATION

android/src/main/java/io/radar/react/RNRadarReceiver.java
tjulien: Hmm, yeah good point.  Do you think this release will cause a regression for https://github.com/radarlabs/react-native-radar/issues/35, or does the fact that the new code path happens not a `BroadcastReciever` make it so this won't happen?

phil-brown: From the stack trace on [the issue](https://github.com/radarlabs/react-native-radar/issues/35), and with the understanding that the broadcast receiver is killed by Android OS if its operation takes too long, it is clear to me that the issue was caused when react had not yet been fully initialized, and the listener was added within the receiver such that its `onReactContextInitialized` callback would be notified when react was ready. This would trigger an event and cause the NPE because the BroadcastReceiver had been garbage collected. I am fairly confident that this is no longer a concern, because the receiver will not be garbage collected by the OS anymore.

android/src/main/java/io/radar/react/RNRadarModule.java
tjulien: cc @jsani-radar

jsani-radar: Out of the 3, my gut says to support 1 and 2 - I'm not sure why someone would use a new helper (#3) when this PR gives the ability for devs to paste a string in an XML file. `Radar.initialize()` is the legacy way of doing things, so I think we should keep that to continue supporting existing apps and avoid a breaking change

phil-brown: The rationale was that the Radar.initialize method does not require any receiver, which makes it so that the hooks into events for RN are not received - so when an existing customer updates to use 3.2, this might easily be missed. Maybe instead of adding RNRadar, it makes sense to do a new minor release number with good docs?

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -129,14 +177,21 @@ public boolean onRequestPermissionsResult(int requestCode, String[] permissions,
...
         if (activity != null) {
-            if (Build.VERSION.SDK_INT >= 23) {
-                if (background && Build.VERSION.SDK_INT >= 29) {
-                    activity.requestPermissions(new String[] { Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION }, PERMISSIONS_REQUEST_CODE, this);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                if (background && Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                    activity.requestPermissions(new String[]{
+                            Manifest.permission.ACCESS_FINE_LOCATION,
+                            Manifest.permission.ACCESS_COARSE_LOCATION,
+                            Manifest.permission.ACCESS_BACKGROUND_LOCATION
```
phil-brown, line 188: Been looking into permissions again, and this BACKGROUND_LOCATION permission needs to be included in a separate permission dialog than FINE/COARSE - Need to make this change before continuing with this PR.

phil-brown: Fixed.

android/src/main/java/io/radar/react/RNRadarModule.java
jsani-radar: ah gotcha, looks like we landed on that approach for [Capacitor]([https://github.com/radarlabs/capacitor-radar/pull/25#discussion_r764265352])

phil-brown: @jsani-radar they will still be able to use 2. I can add back `RNRadar`, giving all 3 options - or we can just have 1 & 2, and have some good documentation for migrating.

android/build.gradle
```diff
@@ -44,6 +45,23 @@ repositories {
 }
 
 dependencies {
+    //noinspection GradleDynamicVersion
```
jsani-radar, line 48: nit: let's have this comment be consistent with the one on L42 - going to start putting together a style guide on comments / release notes

example/android/app/src/release/java/com/example/ReactNativeFlipper.java
```diff
@@ -0,0 +1,16 @@
...
+import com.facebook.react.ReactInstanceManager;
+
+/**
+ * Dummy Flipper initialization. This allows {@link io.radar.example.MainApplication} to call the
+ * {@link #initializeFlipper(Context, ReactInstanceManager)} method without reflection. The debug variant of this
+ * application uses this same method signature to initialize Flipper.
+ */
+public class ReactNativeFlipper {
+  public static void initializeFlipper(Context context, ReactInstanceManager reactInstanceManager) {
+    //Do nothing
```
jsani-radar, line 14: ditto

jsani-radar (commented):
this lgtm at first glance, given last week's debacle I'm going to recuse myself from approving mobile PRs, plus I'd like @tjulien to give the final +1

in the future, it'd be _much_ easier to review this if we added all the linter changes / `checkstyle` as a separate PR, that'd way I'd be able to tell which lines are substantive changes and which ones are just style corrections, it makes reviewing pretty tedious if I'm not sure which changes are important

android/src/main/java/io/radar/react/RNRadarReceiver.java
```diff
@@ -35,51 +39,52 @@ private void sendEvent(final String eventName, final Object data) {
     }
 
     @Override
-    public void onEventsReceived(@NonNull Context context, @NonNull RadarEvent[] events, @Nullable RadarUser user) {
+    public void onEventsReceived(@NonNull Context context,
+                                 @NonNull RadarEvent[] events,
+                                 @Nullable RadarUser user) {
```
jrtibbetts, line 44: 👍

example/android/app/build.gradle
```diff
@@ -78,13 +78,13 @@ import com.android.build.OutputFile
...
+        enableHermes: false
 ]
 
 apply from: '../../node_modules/react-native-unimodules/gradle.groovy'
-apply from: "../../node_modules/react-native/react.gradle"
-apply from: "../../node_modules/expo-constants/scripts/get-app-config-android.gradle"
-apply from: "../../node_modules/expo-updates/scripts/create-manifest-android.gradle"
+apply from: '../../node_modules/react-native/react.gradle'
+apply from: '../../node_modules/expo-constants/scripts/get-app-config-android.gradle'
+apply from: '../../node_modules/expo-updates/scripts/create-manifest-android.gradle'
```
jrtibbetts, line 87: What's the difference between single and double quotes?

example/android/app/src/main/java/io/radar/example/MainApplication.java
```diff
@@ -1,123 +1,88 @@
...
-      } catch (ClassNotFoundException e) {
-        e.printStackTrace();
-      } catch (NoSuchMethodException e) {
-        e.printStackTrace();
-      } catch (IllegalAccessException e) {
-        e.printStackTrace();
-      } catch (InvocationTargetException e) {
-        e.printStackTrace();
-      }
+        //Loads Flipper in React Native templates. Call this in the onCreate method.
```
jrtibbetts, line 85: `// Loads`?

android/build.gradle
phil-brown: @jsani-radar this "noinspection" comment is a keyword used by the linter, and is not a standard comment. There are a few things like this - noinspection, region/end region, etc, that we can call out in the style guide.

example/android/app/src/release/java/com/example/ReactNativeFlipper.java
phil-brown: Fixed

phil-brown:
@tjulien can you review?

jrtibbetts (approved):

PR #159, opened 2021-11-19
react-native-radar  upgrade iOS SDK to 3.2.0

brettguenther:


tjulien:
We'll need to cut a release of iOS (3.2.1 or 3.2.2 to keep consistent with Android) to add support for `CLLocationSourceInformation` :

https://github.com/radarlabs/radar-sdk-ios/compare/3.2.0...master

And then point to that release in this PR.

nickpatrick (approved):

PR #161, opened 2021-12-28
Add Trip Callbacks and Update Trip function

brettguenther:


ios/RNRadar.m
```diff
@@ -314,17 +314,118 @@ - (void)didLogMessage:(NSString *)message {
...
+            rejecter([Radar stringForStatus:status], [Radar stringForStatus:status], nil);
+        }
+        resolver = nil;
+        rejecter = nil;
+    }];
+}
+
+RCT_EXPORT_METHOD(updateTrip:(NSDictionary *)optionsDict resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject) {
+    RadarTripOptions *options = [RadarTripOptions tripOptionsFromDictionary:optionsDict[@"options"]];
+    NSString *statusStr = optionsDict[@"status"];
```
tjulien, line 392: Since we're going to read "status" from the `optionsDict`, we should reject as a bad request if `optionsDict` is null.

Here is a good example of doing that: https://github.com/radarlabs/react-native-radar/blob/master/ios/RNRadar.m#L621

tjulien: I think we should also enforce that people pass a `"status"`, since that's a required param in the underlying implementation call.  If not reject as bad request.

ios/RNRadar.m
```diff
@@ -314,17 +314,118 @@ - (void)didLogMessage:(NSString *)message {
...
+            status = RadarTripStatusStarted;
+        } else if ([statusStr isEqualToString:@"approaching"]) {
+            status = RadarTripStatusApproaching;
+        } else if ([statusStr isEqualToString:@"arrived"]) {
+            status = RadarTripStatusArrived;
+        } else if ([statusStr isEqualToString:@"completed"]) {
+            status = RadarTripStatusCompleted;
+        } else if ([statusStr isEqualToString:@"canceled"]) {
+            status = RadarTripStatusCanceled;
+        }
```
tjulien, line 406: I think we should reject as a bad request if `"status"` does not equal any of the predefined values.

tjulien, line 406: I think we should reject as a bad request here if `"statusStr"` doesn't equal any of the predefined values

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -331,41 +331,158 @@ public void rejectEvent(String eventId) {
...
     }
 
+    @ReactMethod
+    public void updateTrip(ReadableMap optionsMap, final Promise promise) {
+        try {
+            JSONObject optionsObj = RNRadarUtils.jsonForMap(optionsMap);
+            RadarTripOptions options = RadarTripOptions.fromJson(optionsObj.getJSONObject("options"));
+            RadarTrip.RadarTripStatus status = RadarTrip.RadarTripStatus.UNKNOWN;
+
+            if (optionsObj.has("status")) {
```
tjulien, line 437: See the comments in the objective C impl below - we should mirror those same changes here

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -331,41 +331,158 @@ public void rejectEvent(String eventId) {
...
+                        }
+                    } catch (JSONException e) {
+                        Log.e(TAG, "JSONException", e);
+                        promise.reject(Radar.RadarStatus.ERROR_SERVER.toString(), Radar.RadarStatus.ERROR_SERVER.toString());
+                    }
                 }
             });
         } catch (JSONException e) {
             Log.e(TAG, "JSONException", e);
+            promise.reject(Radar.RadarStatus.ERROR_SERVER.toString(), Radar.RadarStatus.ERROR_SERVER.toString());
```
tjulien, line 366: Let's do `Radar.RadarStatus.ERROR_BAD_REQUEST.toString()` for these outer `catch`'es, where the request data is the thing throwing

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -331,41 +331,158 @@ public void rejectEvent(String eventId) {
...
+                        }
+                    } catch (JSONException e) {
+                        Log.e(TAG, "JSONException", e);
+                        promise.reject(Radar.RadarStatus.ERROR_SERVER.toString(), Radar.RadarStatus.ERROR_SERVER.toString());
+                    }
+                }
+            });
+        } catch (JSONException e) {
+            Log.e(TAG, "JSONException", e);
+            promise.reject(Radar.RadarStatus.ERROR_SERVER.toString(), Radar.RadarStatus.ERROR_SERVER.toString());
```
tjulien, line 482: `Radar.RadarStatus.ERROR_BAD_REQUEST.toString()`

tjulien:
@brettguenther Left a few small comments but looks ready otherwise

tjulien:
https://app.shortcut.com/radarlabs/story/8172/add-trip-callbacks-and-update-trip-function-to-react-native

ios/RNRadar.m
brettguenther: i was mirroring what we do on cordova: https://github.com/radarlabs/cordova-plugin-radar/blob/master/src/ios/CDVRadar.m#L353 but agree we should adjust

tjulien (approved):

PR #164, opened 2022-01-04
github action to create react-native-radar SDK releases

tjulien:
As documented here: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

### DIFF
--- a/.github/workflows/radar-release-actions.yml
+++ b/.github/workflows/radar-release-actions.yml
@@ -1,0 +1,18 @@
+name: Publish Package to npmjs
+on:
+  release:
+    types: [created]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm publish
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/radar-release-actions.yml
```diff
@@ -0,0 +1,18 @@
+name: Publish Package to npmjs
+on:
+  release:
+    types: [created]
```
phil-brown, line 4: We've been using `published` instead of `created` here on the Android and iOS GitHub Actions Triggers. May be good to use the same, for consistent behavior when cutting releases.

```suggestion
    types: [ published ]
```

phil-brown (commented):
Awesome! Left one comment. Also, should this build be running any tests, build or validation for the native wrappers? Does `npm ci` and `npm publish` cover that? Looks like the circleci config file also runs `npm run test` and does some native-specific stuff.

tjulien:
@phil-brown It's a good question, right now we're relying on a clean build and test run on `master` before creating the Release. So the tests have to pass, and the additional ios and android build steps have to work otherwise we'll be blocked from merging into `master`.  As an incremental improvement so that we don't have to do releases manually, having the release workflow is probably a good first step. Longer term, we could move the circle build steps into a Github Action (stop using circle), and then that would make it easier to re-use the steps and run them again during the "release" build.

tjulien:
Going to merge this to do some testing

PR #168, opened 2022-02-16
Track once high accuracy and beacons

brettguenther:


example/App.js
```diff
@@ -62,6 +62,27 @@ export default function App() {
     console.log('trackOnce:', err);
   });
 
+  Radar.trackOnce({
```
tjulien, line 65: @gabrielalao Does a plain `trackOnce` with no parameters still work?  Basically, the call right above this one (line 59)

tjulien: Can you run that and paste the output here?

gabrielalao: @tjulien Yes, it's working.

ios/RNRadar.m
```diff
@@ -179,44 +179,25 @@ - (void)didLogMessage:(NSString *)message {
...
-        } else if (rejecter) {
-            rejecter([Radar stringForStatus:status], [Radar stringForStatus:status], nil);
-        }
-        resolver = nil;
-        rejecter = nil;
-    }];
-}
+    NSDictionary *locationDict = optionsDict[@"location"];
 
-RCT_EXPORT_METHOD(trackOnce:(NSDictionary *)locationDict resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject) {
```
tjulien, line 208: So one thing I'm realizing @brettguenther is that the old code here supported this format:
```
Radar.trackOnce({
  latitude: 39.2904,
  longitude: -76.6122,
})
```
And the new code supports nesting `latitude` and `longitude` under `location { ... }` like this:
```
Radar.trackOnce({
  location: {
    latitude: 39.2904,
    longitude: -76.6122,
  }
})
```

So this change would probably break any existing callers.

Wondering if we should keep the old format for backcompat

cc @gabrielalao

gabrielalao: @tjulien We can make our code back compatible. I can update javascript code for handling parameters.

ios/RNRadar.m
```diff
@@ -238,13 +219,32 @@ - (void)didLogMessage:(NSString *)message {
...
+
+    if (accuracy != nil && [accuracy isKindOfClass:[NSString class]]) {
+        if ([accuracy isEqualToString:@"high"] || [accuracy isEqualToString:@"HIGH"]) {
+            desiredAccuracy = RadarTrackingOptionsDesiredAccuracyHigh;
+        } else if ([accuracy isEqualToString:@"medium"] || [accuracy isEqualToString:@"MEDIUM"]) {
+            desiredAccuracy = RadarTrackingOptionsDesiredAccuracyMedium;
+        } else if ([accuracy isEqualToString:@"low"] || [accuracy isEqualToString:@"LOW"]) {
+            desiredAccuracy = RadarTrackingOptionsDesiredAccuracyLow;
+        }
+    }
```
jrtibbetts, line 232: This was probably copied from elsewhere, but it's probably safest to do a case-insensitive comparison, or better yet, lowercase the `accuracy` and compare it to lowercase values.

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -175,56 +175,41 @@ public void onComplete(@NonNull Radar.RadarStatus status, @Nullable Location loc
...
+            if (optionsMap.hasKey("accuracy")) {
+                String accuracy = optionsMap.getString("accuracy");
+                if (accuracy.equals("none") || accuracy.equals("NONE")) {
+                    accuracyLevel = RadarTrackingOptions.RadarTrackingOptionsDesiredAccuracy.NONE;
+                } else if (accuracy.equals("low") || accuracy.equals("LOW")) {
+                    accuracyLevel = RadarTrackingOptions.RadarTrackingOptionsDesiredAccuracy.LOW;
+                } else if (accuracy.equals("medium") || accuracy.equals("MEDIUM")) {
+                    accuracyLevel = RadarTrackingOptions.RadarTrackingOptionsDesiredAccuracy.MEDIUM;
+                } else if (accuracy.equals("high") || accuracy.equals("HIGH")) {
+                    accuracyLevel = RadarTrackingOptions.RadarTrackingOptionsDesiredAccuracy.HIGH;
```
jrtibbetts, line 204: See my other comment about case-insensitive comparisons.

ios/RNRadar.m
```diff
@@ -179,44 +179,25 @@ - (void)didLogMessage:(NSString *)message {
...
-    }];
-}
+    NSDictionary *locationDict = optionsDict[@"location"];
 
-RCT_EXPORT_METHOD(trackOnce:(NSDictionary *)locationDict resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject) {
     CLLocation *location;
     if (locationDict != nil && [locationDict isKindOfClass:[NSDictionary class]]) {
         double latitude = [RCTConvert double:locationDict[@"latitude"]];
         double longitude = [RCTConvert double:locationDict[@"longitude"]];
+        double accuracy = [RCTConvert double:locationDict[@"accuracy"]];
```
jrtibbetts, line 196: Is there/should there be a default value of `5`?

ios/RNRadar.m
```diff
@@ -238,13 +219,32 @@ - (void)didLogMessage:(NSString *)message {
...
+            desiredAccuracy = RadarTrackingOptionsDesiredAccuracyLow;
+        }
+    }
+
+    BOOL beacons = optionsDict[@"beacons"];
+
+    if (beacons) {
+        beaconsTrackingOption = beacons;
+    }
+
```
jrtibbetts, line 239: If `beaconsTrackingOption` isn't used elsewhere, these lines can be moved into the `else` clause below.

jrtibbetts (approved):

ios/RNRadar.m
gabrielalao: accuracy is required field of location object, if it's omitted, it will show the error.

ios/RNRadar.m
gabrielalao: agree

ios/RNRadar.m
jrtibbetts: Ah, ok.

PR #188, opened 2022-06-06
Bump to 3.4.x

nickpatrick:


example/ios/Example.xcodeproj/project.pbxproj
```diff
@@ -303,7 +303,7 @@
 					"FB_SONARKIT_ENABLED=1",
 				);
 				INFOPLIST_FILE = example/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 12.1;
```
jrtibbetts, line 306: Interesting. Any particular reason? The native SDK still supports 10.

jrtibbetts (approved):
You didn't have to do anything with the `RadarRouteMode` declarations?

example/ios/Example.xcodeproj/project.pbxproj
nickpatrick: I was running into a build error that I ended up fixing by updating the Xcode version instead. May revert this.

nickpatrick: > You didn't have to do anything with the `RadarRouteMode` declarations?

Not sure yet. Still trying to repro the original issue and then will make changes if needed.

nickpatrick: Fixed this by moving `RadarRouteMode` to a separate header and bumping versions.

PR #189, opened 2022-06-07
expose setForegroundServiceOptions()

nickpatrick:


ios/RNRadar.m
```diff
@@ -306,6 +306,10 @@ - (void)didLogMessage:(NSString *)message {
     [Radar stopTracking];
 }
 
+RCT_EXPORT_METHOD(setForegroundServiceOptions) {
+    // not implemented
```
jrtibbetts, line 310: Should this throw an error or something until I implement it? Or just leave out these three lines entirely?

jrtibbetts (approved):

ios/RNRadar.m
nickpatrick: It shouldn't throw an error, it can just no-op silently. That way the developer doesn't need to add any platform checks in the cross-platform JS code.

PR #208, opened 2023-01-04
Expose initialize function

lmeier:
For: https://linear.app/radarlabs/issue/MOB-60/expose-initialize-in-the-react-native-sdk

Attempting to pull this into a sample app to test

### DIFF
--- a/android/src/main/java/io/radar/react/RNRadarModule.java
+++ b/android/src/main/java/io/radar/react/RNRadarModule.java
@@ -58,6 +58,11 @@ public class RNRadarModule extends ReactContextBaseJavaModule implements Permiss
     }
 
     @ReactMethod
+    public void initialize(String publishableKey) {
+        Radar.initialize(getReactApplicationContext(), publishableKey);
+    }
+
+    @ReactMethod
     public void setLogLevel(String level) {
         Radar.RadarLogLevel logLevel = Radar.RadarLogLevel.NONE;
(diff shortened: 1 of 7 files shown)


tjulien:
@lmeier There's a sample app in `/example`, that's usually how we do testing of the x-platform SDKs (even just in the simulator can be useful for something like this)

lmeier:
> @lmeier There's a sample app in `/example`, that's usually how we do testing of the x-platform SDKs (even just in the simulator can be useful for something like this)

I wasn't able to get the sample app up & running in this directory. However, I was able to successfully pull in this branch into the expo sample app I've just now got working, and indeed this `initialize` works.

Will be nice to get this in—will allow RN developers to integrate now without touching any native code whatsoever. We can update the docs accordingly.

Finally, you'll see I moved `RadarJS` from a dev dependency to a normal dependency. I think this should have happened in https://github.com/radarlabs/react-native-radar/commit/27569b0ec13ccd796df015fd9bd6fcae3a5ef8a3#diff-eee1687aee979ceaa9debd91891a42c9e8a5b6417ee84c7b5272d6b3cc6eda90 when we added support for web apps.

tjulien:
Nice, I'll test using the example app once this is merged - I have a good local set up for it.

+1 For updating the docs and for bringing this to the other x-platforms.

tjulien (approved):

PR #214, opened 2023-02-27
Fix example app metro config

brettguenther:
Fixes MOB-88.

Our metro configuration for our sample app is currently broken (`Error: Unable to resolve module react-native-radar`). This fixes it.

### DIFF
--- a/example/metro.config.js
+++ b/example/metro.config.js
@@ -1,3 +1,5 @@
+// https://github.com/callstack/react-native-builder-bob
+
 const exclusionList = require('metro-config/src/defaults/exclusionList');
 
 const path = require('path');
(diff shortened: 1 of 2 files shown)


example/metro.config.js
```diff
@@ -1,19 +1,33 @@
-// https://github.com/facebook/react-native/issues/24065#issuecomment-537489786
+const exclusionList = require('metro-config/src/defaults/exclusionList');
```
tjulien, line 2: Did this new `metro.config.js` come from somewhere?  If so, would be help to add a comment with a link to it.

brettguenther: could ref: https://github.com/callstack/react-native-builder-bob/blob/1000cd7c4d99647f4ad2e90cfcc5874697e270aa/packages/create-react-native-library/templates/common/example/metro.config.js

tjulien (approved):

PR #217, opened 2023-03-03
Bump to 3.5.12

lmeier:
iOS to 3.5.12
Android to ~~3.5.13~~ 3.5.12
Web SDK to 3.6.0

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -45,5 +45,5 @@ repositories {
 
 dependencies {
     api 'com.facebook.react:react-native:+'
-    api 'io.radar:sdk:3.5.13'
+    api 'io.radar:sdk:3.5.12'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -18,7 +18,7 @@ android {
         minSdkVersion 16
         targetSdkVersion 31
         versionCode 1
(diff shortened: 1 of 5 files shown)


android/build.gradle
```diff
@@ -18,7 +18,7 @@ android {
         minSdkVersion 16
         targetSdkVersion 31
         versionCode 1
-        versionName '3.5.10'
+        versionName '3.5.11'
```
tjulien, line 21: We should probably set these both to `3.5.12` to keep it consistent with iOS (And watching what's going on with Android, I think there's going to be another release in the next day or two anyway)

android/build.gradle
```diff
@@ -45,5 +45,5 @@ repositories {
 
 dependencies {
     api 'com.facebook.react:react-native:+'
-    api 'io.radar:sdk:3.5.10'
+    api 'io.radar:sdk:3.5.13'
```
tjulien, line 48: We should probably set this to `3.5.12` to keep it consistent with iOS (Also watching what's happening with Android, I think there's going to be another release in a day or two)

lmeier: Makes sense to me! Doing now

tjulien (approved):

PR #219, opened 2023-03-13
CORE-294 react native sdk add conversion event + version bump to 3.7.1

lmeier:
This was enough to get iOS puppy chow working.

Remaining todos:

- [x] make analogous changes for android
- [x] ensure full coverage of the native SDK changes

### DIFF
--- a/android/src/main/java/io/radar/react/RNRadarReceiver.java
+++ b/android/src/main/java/io/radar/react/RNRadarReceiver.java
@@ -26,32 +26,13 @@ public class RNRadarReceiver extends RadarReceiver {
 
     private ReactNativeHost reactNativeHost;
     private static final String TAG = "RNRadarReceiver";
-
-    private int listenerCount = 0;
-    boolean hasListeners = false;
+    protected boolean hasListeners = false;
 
     private void sendEvent(final String eventName, final Object data) {
         final ReactInstanceManager reactInstanceManager = reactNativeHost.getReactInstanceManager();
         ReactContext reactContext = reactInstanceManager.getCurrentReactContext();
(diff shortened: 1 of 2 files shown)


david-goodfellow:
Okay this is looking good to me! Can you verify that it was working on puppychow on both android and ios devices? For iOS, bring me in if needed to test on my device @joeychik-radar :)

android/src/main/java/io/radar/react/RNRadarReceiver.java
```diff
@@ -26,14 +27,34 @@ public class RNRadarReceiver extends RadarReceiver {
...
         final ReactInstanceManager reactInstanceManager = reactNativeHost.getReactInstanceManager();
         ReactContext reactContext = reactInstanceManager.getCurrentReactContext();
-        if (reactContext != null) {
+        if (reactContext != null && hasListeners) {
             reactContext.getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class).emit(eventName, data);
         }
     }
 
+    @ReactMethod
+    public void addListener(String eventName) {
```
david-goodfellow, line 42: Can you provide more context on the adding and removing listeners and how it controls whether events are emitted or not?

joeychik-radar: these are for emitting events from native modules to javascript code. if no event listeners are registered, then we don't emit these events. we get compilation warnings when we dont implement add/remove listener functions
https://reactnative.dev/docs/native-modules-android#sending-events-to-javascript

david-goodfellow (approved):

PR #225, opened 2023-03-28
bump android version to 3.7.2

david-goodfellow:


lmeier:
I think we probably should bump iOS to `3.7.3` here

lmeier (approved):
🚋

PR #226, opened 2023-04-14
trackVerified

lmeier:


nickpatrick:
Let's deal with the Android `3.8` upgrade in a separate PR. I don't expect anyone to try to use `verified` and `replaying` at the same time.

nickpatrick (approved):

nickpatrick:
I updated iOS to `3.7.7` with the completion handler fix, however.

PR #231, opened 2023-06-23
Update scheduledArrivalAt handling on iOS

lmeier:
I initially was fixing with a change to `RadarTripOptions.m`, but figured that isolating the changes to just react native would be best.

I've confirmed the change via Waypoint. It now works with `arrivalTime.getTime()` as we had in the example in this repo, as well as `arrivalTime.toIsoString()` which was undocumented but working on Android.

As far as I can, both Android and iOS handle both datetime formats now, but I'll put through some reps on an Android build over the weekend to triple confirm.

### DIFF
--- a/ios/RNRadar.m
+++ b/ios/RNRadar.m
@@ -415,10 +415,6 @@ RCT_EXPORT_METHOD(startTrip:(NSDictionary *)optionsDict resolve:(RCTPromiseResol
     if (tripOptionsDict == nil) {
         tripOptionsDict = optionsDict;
     }
-    NSObject *scheduledArrivalAtObj = tripOptionsDict[@"scheduledArrivalAt"];
-    if (scheduledArrivalAtObj && [scheduledArrivalAtObj isKindOfClass:[NSNumber class]]) {
-        [tripOptionsDict setObject:[RCTConvert NSDate:scheduledArrivalAtObj] forKey:@"scheduledArrivalAt"];
-    }
 
     RadarTripOptions *options = [RadarTripOptions tripOptionsFromDictionary:tripOptionsDict];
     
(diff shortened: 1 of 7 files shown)


ios/RNRadar.m
```diff
@@ -411,14 +411,17 @@ - (void)didLogMessage:(NSString *)message {
...
+    NSMutableDictionary *tripOptionsDict = optionsDict[@"tripOptions"];
     if (tripOptionsDict == nil) {
         tripOptionsDict = optionsDict;
     }
-    RadarTripOptions *options = [RadarTripOptions tripOptionsFromDictionary:tripOptionsDict];
-    if (options.scheduledArrivalAt) {
-        options.scheduledArrivalAt = [RCTConvert NSDate:options.scheduledArrivalAt];
+    NSObject *scheduledArrivalAtObj = tripOptionsDict[@"scheduledArrivalAt"];
+    if (scheduledArrivalAtObj && [scheduledArrivalAtObj isKindOfClass:[NSNumber class]]) {
+        [tripOptionsDict setObject:[RCTConvert NSDate:scheduledArrivalAtObj] forKey:@"scheduledArrivalAt"];
```
tjulien, line 420: I think your original instinct was good - we do centralize these conversions in the ios and android sdk's so that they're consistent across all platforms.  [Android already supports passing in `long's`](https://github.com/radarlabs/radar-sdk-android/blob/master/sdk/src/main/java/io/radar/sdk/RadarTripOptions.kt#L70), we should probably do the same thing on iOS to keep them consistent.

lmeier: Ok, did that: https://github.com/radarlabs/radar-sdk-ios/pull/264

tjulien (approved):

PR #232, opened 2023-06-25
UI elements

lmeier:


ios/RNRadar.m
```diff
@@ -730,19 +730,16 @@ - (void)didLogMessage:(NSString *)message {
...
-            reject([Radar stringForStatus:RadarStatusErrorBadRequest], [Radar stringForStatus:RadarStatusErrorBadRequest], nil);
-        }
-
-        return;
+    CLLocation *near = nil;
+    if (nearDict && [nearDict isKindOfClass:[NSDictionary class]]) {
+        double latitude = [RCTConvert double:nearDict[@"latitude"]];
+        double longitude = [RCTConvert double:nearDict[@"longitude"]];
+        NSDate *timestamp = [NSDate new];
+        near = [[CLLocation alloc] initWithCoordinate:CLLocationCoordinate2DMake(latitude, longitude) altitude:-1 horizontalAccuracy:5 verticalAccuracy:-1 timestamp:timestamp];
```
david-goodfellow, line 738: just curious, what is the need for setting altitude, horizontalAccuracy, and verticalAccuracy?

js/ui/autocomplete.jsx
```diff
@@ -0,0 +1,148 @@
...
+  TextInput,
+  FlatList,
+  TouchableOpacity,
+  Text,
+  StyleSheet,
+  Image,
+} from 'react-native';
+import Radar from '../index.native';
+
+// import Logger from '../logger'; // Your logger module
```
david-goodfellow, line 14: nit: remove

js/ui/map.jsx
```diff
@@ -0,0 +1,63 @@
...
+  const host = await getHost();
+  const publishableKey = await getPublishableKey();
+  return `${host}/maps/styles/${style}?publishableKey=${publishableKey}`;
+};
+
+const RadarMap = ({ mapOptions, children }) => {
+  const [styleURL, setStyleURL] = useState(null);
+  createStyleURL(mapOptions?.mapStyle || DEFAULT_STYLE).then((result) => {
+    setStyleURL(result);
+  });
```
kochis, line 35: Does this not trigger an infinite loop? It looks like this function is called during each render (and setting the state)

js/ui/map-logo.png
kochis, line 1: Lets use the same logo as web: https://api.radar.io/maps/static/images/logo.svg

js/ui/autocomplete.jsx
```diff
@@ -0,0 +1,209 @@
...
+    (text) => {
+      setQuery(text);
+      if (text.length < config.threshold) {
+        return;
+      }
+      const timer = setTimeout(() => {
+        fetchResults(text);
+      }, config.debounceMS);
+
+      return () => clearTimeout(timer);
```
kochis, line 124: Does the timer get cleared if another input change comes in before the previous one is fired? It seems like this will still fire all the API calls, just after a delay (a new input change should interrupt any pending request, to reduce API call volume)

ios/RNRadar.m
lmeier: It's required for CLLocation: https://developer.apple.com/documentation/corelocation/cllocation/3524339-initwithcoordinate

js/ui/map.jsx
lmeier: Good catch. Fixed: https://github.com/radarlabs/react-native-radar/pull/232/commits/d7b78b6a71c9351bc830e03b619c88dbc37f54ac

js/ui/autocomplete.jsx
lmeier: Ah, indeed we need to cancel previous pending reqs. Here's the fix: https://github.com/radarlabs/react-native-radar/pull/232/commits/3e93f55d15ba2e1bf43609bdb99b94a4b2599fff

js/ui/README.md
```diff
@@ -0,0 +1,67 @@
+## Example usage
+
+We provide UI elements for autocomplete & maps to make building easy.
+
+If you're using the Map element, you'll need to install [Maplibre React Native](https://github.com/maplibre/maplibre-react-native).
```
kochis, line 5: What happens if you don't install it? Does it still compile?

js/ui/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+export default function App() {
+    const [location, setLocation] = useState(null);
+
+    Radar.trackOnce().then((result) => {
+        setLocation({
+          latitude: result.location.latitude,
+          longitude: result.location.longitude,
+        });
+      });
```
kochis, line 53: This should probably be in a `useEffect` block, otherwise it runs each render

js/ui/README.md
```diff
@@ -0,0 +1,157 @@
...
+  }, []);
+
+  const onSelect = (selectedAddress) => {
+      // Do something with the selected address
+  }
+
+  return (
+      <View style={{ marginTop: 50}}>>
```
brettguenther, line 35: extra > here

js/ui/map.jsx
```diff
@@ -0,0 +1,76 @@
+import React, { useState, useEffect } from 'react';
+import { StyleSheet, View, Image } from 'react-native';
+import { getHost, getPublishableKey } from '../helpers';
+
+let MapLibreGL;
+try {
+  MapLibreGL = require('@maplibre/maplibre-react-native');
```
kochis, line 7: I'm surprised it lets you do this using ES modules 🤔

kochis, line 7: I'm surprised it lets you do this using ES modules 🤔

js/ui/map.jsx
```diff
@@ -0,0 +1,76 @@
...
+    left: 5,
+    width: 50,
+    height: 50,
+    resizeMode: 'contain',
+  },
+});
+
+const createStyleURL = async (style = DEFAULT_STYLE) => {
+  const host = await getHost();
+  const publishableKey = await getPublishableKey();
```
kochis, line 33: Nit, can be run concurrently:
```js
const [host, publishableKey] = await Promise.all([getHost(), getPublishableKey()]);
```

js/ui/map.jsx
kochis: Ah, everything is compiled to CommonJS: https://stackoverflow.com/a/46336345/631816

kochis (approved):

PR #238, opened 2023-07-25
Fix github release process

lmeier:
Presently our github actions release pre-releases as `latest`. These changes ensure that beta / pre-releases get marked as such so that `npm install react-native-radar` only installs the latest, non beta versions.

### DIFF
--- a/.github/workflows/radar-pre-release-actions.yml
+++ b/.github/workflows/radar-pre-release-actions.yml
@@ -12,8 +12,8 @@ jobs:
         with:
           node-version: '16.x'
           registry-url: 'https://registry.npmjs.org'
+      - run: npm ci --force
       - run: npm run check-beta-tag
-      - run: npm ci
       - run: npm publish --tag beta
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/radar-pre-release-actions.yml
+++ b/.github/workflows/radar-pre-release-actions.yml
@@ -12,7 +12,7 @@ jobs:
         with:
           node-version: '16.x'
           registry-url: 'https://registry.npmjs.org'
-      - run: npm ci --force
+      - run: npm ci
       - run: npm run check-beta-tag
       - run: npm publish --tag beta
         env:
--- a/.github/workflows/radar-pre-release-actions.yml
+++ b/.github/workflows/radar-pre-release-actions.yml
@@ -1,7 +1,7 @@
 name: Publish Package to npmjs
 on:
   release:
-    types: [ released ]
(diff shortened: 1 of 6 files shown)


scripts/check-latest-tag.cjs
```diff
@@ -0,0 +1,25 @@
+const fs = require('fs');
+const packageJSON = require('../package.json');
+const packageLockJSON = require('../package-lock.json');
+
+let tagVersion = process.env.GITHUB_REF_NAME || '';
+console.log('Checking tag:', tagVersion);
+
+if (!tagVersion.startsWith('v')) {
```
tjulien, line 8: seems like we haven't been prefixing `v`?

https://github.com/radarlabs/react-native-radar/releases

lmeier: Ahhh, removed from the beta check but not from the latest check. Good catch.

https://github.com/radarlabs/react-native-radar/pull/238/commits/517e2f48e29f6fc7a1596b83736dac4022b34e21

.github/workflows/radar-pre-release-actions.yml
```diff
@@ -0,0 +1,19 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci --force
```
kochis, line 15: Do you know if `--force` is required here? We should probably remove it unless it's needed (it's a holdover from an old version of rollup, I need to remove it from the JS SDK as well)

lmeier: Oh, no I don't. I added it after seeing it in the web sdk. I'll remove.

package.json
```diff
@@ -3,7 +3,7 @@
   "description": "React Native module for Radar, the leading geofencing and location tracking platform",
   "homepage": "https://radar.com",
   "license": "Apache-2.0",
-  "version": "3.7.5",
+  "version": "3.7.6",
```
tjulien, line 6: A few more versions to bump, but I assume you're waiting on the android release first.

https://www.notion.so/radarlabs/How-to-release-the-React-Native-SDK-21c70f16e0bd4fe28425923645cadda0

lmeier: Yea — there's now 3 PRs open for RN that we're trying to get in. @nickpatrick 's native version bump, this, and the UI kit. Was just gonna handle the GH workflow stuff in this given all the stuff in flight.

tjulien (approved):

PR #239, opened 2023-07-26
bump versions

nickpatrick:


lmeier (commented):
need to `npm install` to bump the `package-lock.json`

PR #245, opened 2023-08-07
Support using the "style" field in map options

kochis:
* Uses the `style` field in mapOptions.


kochis:
Actually, digging into the MapLibre RN docs a bit more, I think `mapStyle` makes more sense as the param name here (`style` is reserved for styling the map container). Will need to think through how we document this difference.

PR #247, opened 2023-08-10
Add expand units field

KennyHuRadar:
QA:
methodology: use waypoint's call of auto-complete and manually set expandUnits field to be true and false and observe result
image

IOS, when field is set to true, we can see that the SDK includes it in the http request
image

Android, when the field is set to true and the layers is not set, we get a bad request error as expected, setting the field to false subsequently results in a normal call
image


test/index.test.js
```diff
@@ -366,11 +366,14 @@ describe('calls native implementation', () => {
   test('autocomplete', () => {
     const options = {
       query: 'brooklyn roasting',
+      layers: ['"address'],
```
lmeier, line 369: quotation mark malformation here

ios/RNRadar.m
```diff
@@ -787,10 +787,16 @@ - (void)didLogMessage:(NSString *)message {
         country = optionsDict[@"country"];
     }
 
+    BOOL expandUnits = false;
+    NSNumber *expandUnitsPointer = optionsDict[@"expandUnits"];
+    if (expandUnitsPointer != nil && [expandUnitsPointer isKindOfClass:[NSNumber class]]) {
+        expandUnits = [expandUnitsPointer boolValue]; 
+    }
```
lmeier, line 794: look at `RadarUser.m` or a few lines above for more idiomatic naming. Probably want to change `expandUnitsPointer` to `expandUnitsNumber` (or `expandUnitsObj`).

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -915,7 +915,9 @@ public void autocomplete(ReadableMap optionsMap, final Promise promise) {
         String country = optionsMap.hasKey("countryCode") ? optionsMap.getString("countryCode") : optionsMap.hasKey("country") ? optionsMap.getString("country") : null;
         String[] layers = optionsMap.hasKey("layers") ? RNRadarUtils.stringArrayForArray(optionsMap.getArray("layers")) : null;
 
-        Radar.autocomplete(query, near, layers, limit, country, new Radar.RadarGeocodeCallback() {
+        boolean expandUnits = optionsMap.hasKey("expandUnits")? optionsMap.getBoolean("expandUnits") : false;
```
lmeier, line 918: Missing space in `("expandUnits")?`

example/App.js
```diff
@@ -29,6 +29,7 @@ Radar.on("log", (result) => {
 });
 
 export default function App() {
+
```
lmeier, line 32: 🤔

lmeier (commented):
few pieces of feedback

lmeier (approved):
lgtm

PR #250, opened 2023-08-14
bump Android version

nickpatrick:


ios/RNRadar.m
```diff
@@ -91,7 +91,7 @@ - (void)didLogMessage:(NSString *)message {
     }
 }
 
-RCT_EXPORT_METHOD(initialize:(NSString *)publishableKey) {
+RCT_EXPORT_METHOD(initialize:(NSString *)publishableKey fraud:(BOOL)fraud) {
```
nickpatrick, line 94: no iOS equivalent

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -78,9 +78,13 @@ public String getName() {
     }
 
     @ReactMethod
-    public void initialize(String publishableKey) {
-        Radar.initialize(getReactApplicationContext(), publishableKey);
-        Radar.setReceiver(receiver);
+    public void initialize(String publishableKey, boolean fraud) {
+        if (fraud) {
```
nickpatrick, line 82: only changes `initialize()` logic if `fraud == true`

android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -78,9 +78,13 @@ public String getName() {
...
+            Radar.initialize(getReactApplicationContext(), publishableKey, receiver, Radar.RadarLocationServicesProvider.GOOGLE, fraud);
+        } else {
+            Radar.initialize(getReactApplicationContext(), publishableKey);
+            Radar.setReceiver(receiver);
```
lmeier, line 86: Hmm, unless I'm missing something, I think we want to set the receiver regardless of whether or not `fraud: true`

nickpatrick: I could remove the `if/else` and collapse this to `Radar.initialize(getReactApplicationContext(), publishableKey, receiver, Radar.RadarLocationServicesProvider.GOOGLE, fraud)`. Just didn't want to touch existing `initialize()` logic if not necessary.

example/App.js
```diff
@@ -39,7 +39,7 @@ export default function App() {
   const stringify = (obj) => JSON.stringify(obj, null, 2);
 
   useEffect(() => {
-    Radar.initialize("prj_test_pk_4ce462a534c0f749af3fdaac6fff9f485d65e720");
+    Radar.initialize("prj_test_pk_4ce462a534c0f749af3fdaac6fff9f485d65e720", true);
```
lmeier, line 42: Let's change this key back to `prj_test_pk_0000000000000000000000000000000000000000`. A test key accidentally made it here (now invalidated).

lmeier (approved):
noted 2 small things. apart from those, if it's building correctly and testing well for you, this is gtg 🚀

android/src/main/java/io/radar/react/RNRadarModule.java
lmeier: Way of doing that while also setting the receiver if fraud is true would be:
```
if (fraud) {
    Radar.initialize(getReactApplicationContext(), publishableKey, receiver, 
    Radar.RadarLocationServicesProvider.GOOGLE, fraud);
Radar.setReceiver(receiver);
} else { 
    Radar.initialize(getReactApplicationContext(), publishableKey);
    Radar.setReceiver(receiver);
}

File: package.json [hunk 5]
```diff
@@ -3,7 +3,7 @@
   "description": "React Native module for Radar, the leading geofencing and location tracking platform",
   "homepage": "https://radar.com",
   "license": "Apache-2.0",
-  "version": "3.8.4",
+  "version": "3.8.5",
```
tjulien, line 6: `npm install` to update package-lock.json

android/build.gradle
```diff
@@ -45,5 +45,5 @@ repositories {
 
 dependencies {
     api 'com.facebook.react:react-native:+'
-    api 'io.radar:sdk:3.8.5'
+    api 'io.radar:sdk:3.8.6'
```
tjulien, line 48: above: `versionName '3.8.5'`

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: Do we think the separate `setReceiver()` call is needed if `receiver` is passed to `initialize()`?

package.json
nickpatrick: Done

android/src/main/java/io/radar/react/RNRadarModule.java
nickpatrick: Cool

PR #251, opened 2023-08-18
Fix initialize

lmeier:
Makes the `3.8.5` release non-breaking


nickpatrick:
Whoops, good catch. I can take this out this morning

nickpatrick (approved):

PR #257, opened 2023-09-05
Shift when Radar.initialize is called in example app

KennyHuRadar:
`Radar.initialize` is moved from being called inside useEffect to outside of App to prevent Map from trying to load before `Radar.initialize` is called


lmeier (approved):
lgtm!

cc @nickpatrick

PR #258, opened 2023-09-12
Catch Map element possible permissions error

lmeier:
The map element requests location for the location indicator, but we can't be sure that we have the necessary permissions.

The `react-native-radar` sdk doesn't do full permission monitoring, so the safest thing to do is to attempt to get the location and fail gracefully.

I'm presently `console.warn`-ing, but we probably should replicate the [logger](https://github.com/radarlabs/radar-sdk-js/blob/master/src/logger.ts) used in the js sdk sometime soon.

### DIFF
--- a/js/ui/map.jsx
+++ b/js/ui/map.jsx
@@ -31,13 +31,12 @@ const RadarMap = ({ mapOptions, children }) => {
 
   useEffect(() => {
     Radar.getLocation().then((result) => {
-      if (!result.location || !result.location.latitude || !result.location.longitude) {
-        return;
+      if (result?.location?.latitude && result?.location?.longitude) {
+        setUserLocation({
+          latitude: result.location.latitude,
+          longitude: result.location.longitude,
+        });
       }
-      setUserLocation({
-        latitude: result.location.latitude,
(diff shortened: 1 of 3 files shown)


js/ui/map.jsx
```diff
@@ -31,10 +31,16 @@ const RadarMap = ({ mapOptions, children }) => {
 
   useEffect(() => {
     Radar.getLocation().then((result) => {
+      if (!result.location || !result.location.latitude || !result.location.longitude) {
```
kochis, line 34: Is it possible for result to be null?
```js
if (!result || !result.location || !result.location.latitude || !result.location.longitude) {
```

kochis (approved):

js/ui/map.jsx
lmeier: yes, if the status isn't `RadarStatusSuccess`

kochis: Might be simpler to just do:
```js
if (result?.location?.latitude && result?.location?.longitude) {
  setUserLocation({
    latitude: result.location.latitude,
    longitude: result.location.longitude,
   });
}
```

PR #262, opened 2023-10-09
implemented endpoints for getting versioning for RN and native

KennyHuRadar:
This PR exposes endpoints so that RN SDK can easily access the versioning of the RN SDK and also the underlying native SDK from code.

There are 2 new endpoints:

- `nativeSdkVersion` 
- `rnSdkVersion`

A notable change is that the explicit definition of the versioning of the Android native SDK had been moved from `build.gradle` to `gradle.property`

QA:
we call both of these endpoints in waypoint to determine the versioning of the dependencies 

Android:
image
IOS:
image

### DIFF
--- a/js/index.native.js
+++ b/js/index.native.js
@@ -20,7 +20,7 @@ const setUserId = (userId) => {
 };
 
 const getUserId = () => (
-  NativeModules.RNRadar.getUserId()206G
+  NativeModules.RNRadar.getUserId()
 );
 
 const setDescription = (description) => {
--- a/js/index.native.js
+++ b/js/index.native.js
@@ -205,7 +205,6 @@ const nativeSdkVersion = () => (
 
 const rnSdkVersion = () => (version)
 
-
 const Radar = {
   initialize,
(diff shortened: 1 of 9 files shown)


js/index.native.js
```diff
@@ -1,4 +1,5 @@
 import { NativeEventEmitter, NativeModules, Platform } from 'react-native';
+import {version} from '../package.json'
```
lmeier, line 2: Need a single space b/t each side of `version` and the curlies

js/index.native.js
```diff
@@ -198,6 +199,13 @@ const off = (event, callback) => {
   }
 };
 
+const nativeSdkVersion = () => (
+  NativeModules.RNRadar.nativeSdkVersion()
+);
+
+const rnSdkVersion =()=>(version)
```
lmeier, line 206: spacing

lmeier (commented):
some nits, we'll also need an iOS bump right?

lmeier (approved):
🚀

PR #263, opened 2023-10-30
Start tracking verified ios only

KennyHuRadar:
This PR implements the RN bridge for startTrackingVerified and the event listener for the 'token' event.

QA in Waypoint:
image
We can successfully get back a token from the server, which then can be successfully decoded


ios/RNRadar.m
```diff
@@ -91,6 +91,12 @@ - (void)didLogMessage:(NSString *)message {
     }
 }
 
+- (void)didUpdateToken:(NSString *)token {
+    if(hasListeners){
```
lmeier, line 95: spacing

PR #269, opened 2023-11-10
Bump version to 3.8.14, update Radar SDK to 3.8.15

KennyHuRadar:


lmeier (approved):
Wait for it to hit gradle and then this is good

PR #270, opened 2023-11-13
Wrapper for setAndroidNotificationOptions

KennyHuRadar:
- Wrapper for setAndroidNotification settings
- Bump version
- IOS wrapper for `setForegroundOptions` and `setAndroidNotificationOptions` now takes a dictionary such that they are truly no-ops rather than crashing upon called on ios.


android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -527,6 +528,17 @@ public void setForegroundServiceOptions(ReadableMap optionsMap) {
         }
     }
 
+    @ReactMethod
+    public void setAndroidNotificationsOptions(ReadableMap optionsMap) {
```
lmeier, line 532: make this singular

also probably should make this just be `setNotificationOptions` — we might eventually have an iOS one and it'd be nice to make them the same (will probably need to be clever in how we handle some params)

lmeier (approved):

PR #278, opened 2023-12-19
Bump version to 3.8.18, iOS to 3.8.12

lmeier:


lmeier:
> I think you forgot to add bump the version in the gradle file. The `versionName` need to be bumped regardless.

There's no corresponding android bump necessary, but correct re: `versionName`, good catch, just fixed.

KennyHuRadar (approved):
looks good

PR #280, opened 2024-01-09
expandUnits -> mailable

york-wei:
This PR changes the `expandUnits` param for autocomplete to `mailable`. Value of `true` is passed in for Android wrapper since there are no default value params in Java. Verified by testing using Waypoint


lmeier (approved):
lgtm

PR #281, opened 2024-01-17
threshold -> minCharacters

york-wei:


KennyHuRadar:
Is there a way we can QA this? Perhaps at the minimum some manual tests that it works and nothing is obviously broken?
@york-wei  if you do a manual QA a screenshot and description of the test would do.

york-wei:
> Is there a way we can QA this? Perhaps at the minimum some manual tests that it works and nothing is obviously broken? @york-wei if you do a manual QA a screenshot and description of the test would do.

@KennyHuRadar I did a test with the example app, added the autocomplete component and tested with different values for `options`

lmeier (approved):
lgtm, only thing is that we should add a `MIGRATIONS.md` file that we can refer to in the release notes here, providing an explicit migration snippet

PR #292, opened 2024-02-12
Update param parsing

york-wei:


ios/RNRadar.m
```diff
@@ -329,19 +335,17 @@ - (void)didUpdateToken:(NSString *)token {
         rejecter = nil;
     };
 
-    if (optionsDict != nil) {
-        BOOL beacons = optionsDict[@"beacons"];
```
york-wei, line 333: This was evaluating to `YES` for all non-nil value passed in for `beacons`

lmeier (approved):
👍

KennyHuRadar (changes requested):
Great catch, that explain the bugs we are getting in waypoint, I noticed that the `trackOnce` calls also uses the same buggy way to parse the `beacons` param, I think we can fix it here as well.

KennyHuRadar (approved):
nice

PR #293, opened 2024-02-14
Set x-platform info on initialize

york-wei:
- Tested on Waypoint


lmeier (approved):
code is good, just need to check and update versions given the bumps

PR #297, opened 2024-02-28
Typing fixes

KennyHuRadar:
- Make `Radar.initialize` have optional second param.
- Add `iconString` and `iconColor` fields to `RadarTrackingOptionsForegroundService`


lmeier (approved):
yea, good catch. we should get this in soon

PR #298, opened 2024-02-29
added presets for tracking options

KennyHuRadar:


lmeier:
Hmmm, is there a way where we can present a union of the android and iOS. Would be best if people code do something like:
```
Radar.startTrackingCustom( { ...presetContinuous, desiredSyncInterval: 5 });
```
without having to split b/t iOS and Android.

lmeier (approved):
nice! We'll definitely add this to the docs, probably should add to the example app as well

PR #300, opened 2024-03-06
Expo plugin

KennyHuRadar:
This PR introduces the following:

1. expo plugins for `react-native-radar`
2. example app that uses continuous native generation that tests correctness of expo plugin. Example app has been bumped to expo SDK 50. 
3. updated CI that checks correctness of plugins. 

Note: The new example app opts to comment out the code for map libre's dependency and the map. This is due to the fact that there are some issues with the current map-libre's expo plugin. We are opting for our CI to test our code and expo plugin instead of map-libre's. That said, care should be taken when we are updating our RN map UI export and manual QA should be undertaken. 

Docs: https://github.com/radarlabs/documentation/pull/451

For this review, there seems to be a lot of lines of code changed but they are mostly auto-generated code from a new expo example app. The files that were modified within the example app are the `package.json`, `app.json`, `App.tsx` and `exampleButton.js`


coreybrown89:
@KennyHuRadar do you know if this will be merged anytime soon?

KennyHuRadar:
> @KennyHuRadar do you know if this will be merged anytime soon?

Yeah this PR did kind of get buried, @lmeier lets revisit this when you get back.

package.json
```diff
@@ -6,20 +6,23 @@
   "version": "3.12.2",
   "main": "dist/src/index.js",
```
ShiCheng-Lu, line 7: for me, tsc build outputted everything under dist directly, so this main entry point should be `dist/index.js` unless there is a version mismatch for where tsc output.
Screenshot 2024-07-02 at 11 17 48 AM

.github/workflows/radar-release-actions.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       - run: npm run check-latest-tag
       - run: npm run build
       - run: npm run copy-assets
+      - run: npm run build-plugin
```
ShiCheng-Lu, line 19: we can probably just run `build-all` here

ShiCheng-Lu:
If we are using expo managed build, we could consider removing android/ios in example app from git. Or we can keep it in as an example of what those files should look like for none-managed projects. (We can only track a subset of the files that are important, like the main .java/kt, Objective-C, manifests, and leave the image assets out)

KennyHuRadar:
> If we are using expo managed build, we could consider removing android/ios in example app from git. Or we can keep it in as an example of what those files should look like for none-managed projects. (We can only track a subset of the files that are important, like the main .java/kt, Objective-C, manifests, and leave the image assets out)

A managed workflow will indeed relegate those native code to auto-generated code.

Flip side is that apps in the wild may be running without expo or use expo's bare workflow like we were previously doing. Removing our native folders may leave them out in the cold.

ShiCheng-Lu:
verified working with local install, we can make a pre-release just in case.

ShiCheng-Lu (approved):

example/package.json
```diff
@@ -1,44 +1,27 @@
...
-    "start": "expo start --dev-client",
-    "android": "react-native run-android",
-    "ios": "react-native run-ios",
+    "start": "expo start",
+    "android": "expo run:android",
+    "ios": "expo run:ios",
     "web": "expo start --web",
-    "eject": "expo eject",
-    "postinstall": "patch-package"
+    "install-radar": "npm un --no-save react-native-radar ; npm i --no-save --install-links",
```
ShiCheng-Lu, line 10: I don't think --no-save works here without having react-native-radar in the dependencies, so just rm --no-save, and let dependencies be "react-native-radar": "file:.."

ShiCheng-Lu: builds fine, but doesn't resolve RNRadar after the app opens, might just be me though

KennyHuRadar: I don't think this is fine. The app building fine usually suggests that the native side stuff are in order. Usually js level issues are encountered at run time. I think changing the entry point to a path that does not exist is a bad idea.

example/ios/Podfile
```diff
@@ -1,43 +1,85 @@
...
-    # Workaround `Cycle inside FBReactNativeSpec` error for react-native 0.64
-    # Reference: https://github.com/software-mansion/react-native-screens/issues/842#issuecomment-812543933
-    installer.pods_project.targets.each do |target|
-      if (target.name&.eql?('FBReactNativeSpec'))
-        target.build_phases.each do |build_phase|
-          if (build_phase.respond_to?(:name) && build_phase.name.eql?('[CP-User] Generate Specs'))
-            target.build_phases.move(build_phase, 0)
-          end
+# @generated begin @maplibre/maplibre-react-native-post_installer - expo prebuild (DO NOT MODIFY) sync-72b8976cc2231a4441a5b54389fb6e10bd42a1be
+    $RCTMLN.post_install(installer)
```
lmeier, line 58: Maybe callout that this is how you do it if you're on maplibre 10 or higher.

Is there any automatic podfile modifications we can do with the expo plugin? Or will we still need to mention to add this step?

lmeier (commented):
A few questions:
- should we add maplibre setup instructions (or a pointer to theirs?) to the docs PR? https://github.com/radarlabs/documentation/pull/451/files
- what's the size diff on this PR? I'm also drawn to ShiCheng's suggestion that we potentially don't include the example apps native code in this—it feels like bloat. That shouldn't affect using the SDK at all.
- some dependencies are now captured in this (e.g. the playstore integrity dep) — let's add some instructions to the react-native-sdk checklist about what potentially needs to be checked for that

example/ios/Podfile
ShiCheng-Lu: we should add the maplibre plugin as well to the docs there (https://github.com/radarlabs/documentation/pull/451/files ) though (and make a comment that it's optional if they are not using maps)

KennyHuRadar: cool, made a note in the docs.

KennyHuRadar: @lmeier I think we can omit the native file, the only downside is that people people without expo will no longer be able to build our native folder and hence not be able to run our example app out of the box.

KennyHuRadar: in terms of size diff, its honestly much smaller than it looks,
we are mostly

- changing the CI
- modifying the example app slightly to use expo and continued native genreation
- implementing plugins.

most of the diff is just removing files.

lmeier: image
Looks like it's a 3mb difference, almost 20%.

lmeier: note: it's actually 3.6mb, smaller than the previous 3.9mb. we're fine on this front

lmeier (approved):

ChromeQ:
Will it be a requirement to add this plugin to the expo config?
```json
"plugins": [
      [
        "react-native-radar",
        {
          "iosFraud": true,
          "iosNSLocationAlwaysAndWhenInUseUsageDescription": "test value",
          "iosBackgroundMode": true,
          "androidFraud": true,
          "androidBackgroundPermission": true,
          "androidFineLocationPermission": true
        }
      ],
      ["@maplibre/maplibre-react-native"]
    ]
```

Perhaps some documentation would be good for these options (and defaults) and whether maplibre is a requirement too?
Thanks

KennyHuRadar:
> Will it be a requirement to add this plugin to the expo config?
> 
> ```json
> "plugins": [
>       [
>         "react-native-radar",
>         {
>           "iosFraud": true,
>           "iosNSLocationAlwaysAndWhenInUseUsageDescription": "test value",
>           "iosBackgroundMode": true,
>           "androidFraud": true,
>           "androidBackgroundPermission": true,
>           "androidFineLocationPermission": true
>         }
>       ],
>       ["@maplibre/maplibre-react-native"]
>     ]
> ```
> 
> Perhaps some documentation would be good for these options (and defaults) and whether maplibre is a requirement too? Thanks

No it will not be, it is only needed if you are also using Radar Maps. We have a documentation PR that will go live once we include this change in an upcoming release.

PR #307, opened 2024-03-25
Fix for JS projects

KennyHuRadar:
The previous upgrade to TS introduced a bug that affected projects on older versions of expo using JS. It appears that just loading the import of `radar-js-sdk` into the js layer was causing issues. This PR introduces a fix to prevent that from occuring.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 import { RadarNativeInterface } from "./@types/RadarNativeInterface";
-import {Platform} from "react-native";
+import { Platform } from "react-native";
 
 let module: RadarNativeInterface;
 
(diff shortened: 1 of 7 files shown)


lmeier (commented):
Did the customer running into the issue verify this fix resolved the issue for them?

KennyHuRadar:
> Did the customer running into the issue verify this fix resolved the issue for them?

I cut a beta and send it to them, no replies yet. The QA that was done was that I somewhat recreated a similar expo project and reproduced the error they send via screenshot. This PR fixes that error.

src/index.ts
```diff
@@ -1,12 +1,16 @@
 import { RadarNativeInterface } from "./@types/RadarNativeInterface";
+import {Platform} from "react-native";
```
lmeier, line 2: nit: curly bracket spacing

src/index.ts
```diff
@@ -1,12 +1,16 @@
...
 
 let module: RadarNativeInterface;
 
 module = require("./index.native").default;
 
 export default module;
 
-export { default as RadarRNWeb } from "./index.web";
+let RadarRNWeb =  Platform.OS === 'web'?  require("./index.web").default: {};
```
lmeier, line 10: Hmmm, I'd feel better about this solve if we had a RN Web project to try it on. Basically for all the projects we're aware of we're just removing the import.

tsconfig.json
```diff
@@ -22,7 +22,8 @@
       "noImplicitAny": false,
     },
     "include": [
-      "src/index.ts"
+      "src/index.ts",
+      "src/index.web.js"
```
lmeier, line 26: Hmmm, does. this line alone not solve it?

src/index.ts
lmeier: there's also an extra space before 'Platform'

KennyHuRadar:
I ran a simple example RN web app with the new changes, seems to work fine.
image

lmeier (approved):
nice. let's properly version & ship.

PR #326, opened 2024-06-21
also return foreground granted when coarse location is granted

ShiCheng-Lu:


KennyHuRadar:
we should not ship this, instead we should be writing the RN wrapper for our new location permission handling component.

KennyHuRadar:
Ah I see, I misunderstood the rationale.

KennyHuRadar (approved):

PR #332, opened 2024-06-24
add missing fields to RadarFraud and bump version

ShiCheng-Lu:


ios/RNRadar.m
```diff
@@ -357,7 +357,7 @@ - (void)didUpdateLocationPermissionStatus:(RadarLocationPermissionStatus *)statu
             NSMutableDictionary *dict = [NSMutableDictionary new];
             [dict setObject:[Radar stringForStatus:status] forKey:@"status"];
             if (token != nil) {
-                [dict setObject:[token dictionaryValue] forKey:@"token"];
+                [dict setObject:[RNRadar mapLocationPermissionStatus:[token dictionaryValue]] forKey:@"token"];
```
KennyHuRadar, line 360: can you add a comment explaining this line, its not evident why we are doing this at first glance

ShiCheng-Lu: ios returns the enum as SnakeCaseString, android returns it as CAP_WITH_UNDERSCORE, just making them the same for react-native. I'll add a comment in code.

ShiCheng-Lu: this function is at the top of the file.

KennyHuRadar (approved):

PR #335, opened 2024-06-27
should not call RNRadar.mapLocationPermissionStatus

ShiCheng-Lu:


KennyHuRadar (approved):
We can merge it and just silently ship this with the next release. It's currently just doing nothing so no rush to ship.

PR #338, opened 2024-07-02
Expo plugin maps

ShiCheng-Lu:
all the diffs are in package.lock + Podfile.lock, so it's actually not that scary

### DIFF
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -143,6 +143,10 @@ android {
 
 dependencies {
 
+
+
+
+
     implementation "com.google.android.gms:play-services-location:21.0.1"
 
     implementation "com.google.android.play:integrity:1.2.0"
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -147,6 +147,8 @@ dependencies {
 
 
 
+
+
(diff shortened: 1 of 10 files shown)


example/android/app/build.gradle
```diff
@@ -143,6 +143,12 @@ android {
 
 dependencies {
 
+
+
+
+
+
+
```
ShiCheng-Lu, line 151: expo prebuild just generated a bunch of empty lines... I we leave this here otherwise every build will have this diff?

package.json
```diff
@@ -37,7 +37,7 @@
     ]
   },
   "peerDependencies": {
-    "@maplibre/maplibre-react-native": "^9.0.1",
+    "@maplibre/maplibre-react-native": ">=9.0.1 || >=10.0.0-alpha",
```
ShiCheng-Lu, line 40: peer dependency to any maplibre that is above 9.0.1, alphas will only match exact, so to use 10.0.0-alpha.X, the second part needs to be there.

example/android/app/build.gradle
KennyHuRadar: yeah that's fine, this is basically autogenerated code.

KennyHuRadar (approved):

PR #351, opened 2024-09-12
bump version

KennyHuRadar:


ios/RNRadar.m
lmeier, line 1: We need to handle the new initialize options right?

KennyHuRadar: We still want to call the old method that does not have the automatic setup if we are initializing via the react native layer. This is because we will be initializing after app has launched. We need to manually call [Radar nativeSetup] inside the AppDelegate instead.

lmeier (approved):

PR #352, opened 2024-09-23
Test compatibility with new architecture 

KennyHuRadar:
- A lot of line of code change, but it is mostly changes to the packages and dependencies.
- Upgrades the example app to higher expo version.
- Upgrades the SDK to higher RN version.
- Some tidy up to the example app and fixes to android native code.
- Fixes to jest mocks and tests (along with jest version bump).

This is not us upgrading to the new architecture, more of a compatibility band aid.

We are relying on the interop layer to make the UI element work. This will be removed in the future (according to react native) and we will need to make the UI element migration. We can consider making the change once maplibre updates as well. Right now we are blocked as they are incompatible with the new architecture. 

We are still using nativemodule and not turbomodule, we do not need to perform the migration now but we need to do that sometime in the future. We should then implement a backwards compatible library powered by a backwards compatible turbomodule.


test/jest.setup.js
```diff
@@ -0,0 +1,63 @@
...
+      getUserId: jest.fn(),
+      getDescription: jest.fn(),
+      getMetadata: jest.fn(),
+      setAnonymousTrackingEnabled: jest.fn(),
+      isTracking: jest.fn(),
+      getTrackingOptions: jest.fn(),
+      getTripOptions: jest.fn(),
+      logConversion: jest.fn(),
+      requestForegroundLocationPermission: jest.fn(),
+      requestBackgroundLocationPermission: jest.fn(),
```
lmeier, line 59: Old permissions stuff

lmeier (approved):

PR #354, opened 2024-10-04
patch for conversion revenue

KennyHuRadar:


ios/RNRadar.m
```diff
@@ -1152,9 +1152,9 @@ - (void)didUpdateToken:(RadarVerifiedLocationToken *)token  {
     };
     
     if (revenue) {
-        [Radar logConversionWithName:name metadata:metadata completionHandler:completionHandler];
-    } else {
         [Radar logConversionWithName:name revenue:revenue metadata:metadata completionHandler:completionHandler];
+    } else {
```
lmeier, line 1156: When we're parsing revenue, let's do this:
```
    NSNumber *revenue = nil;
    if (revenueObj != nil) {
        if ([revenueObj isKindOfClass:[NSNumber class]]) {
            revenue = (NSNumber *)revenueObj;
        } else if ([revenueObj isKindOfClass:[NSString class]]) {
            revenue = @([(NSString *)revenueObj doubleValue]);
        }
    }
  ```

and then change `if (revenue) {` to `if (revenue != nil) {`

lmeier (changes requested):

ios/RNRadar.m
KennyHuRadar: we don't need to handle the case where it is a string as the ts is types as number.

KennyHuRadar: image
image

Dogfooding to ensure that the current implementation handles having revenue defined as param or missing in param

lmeier (approved):
let's bump native SDKs as well, then this is good to go

PR #359, opened 2024-10-29
add address validation for web, add other functions

ShiCheng-Lu:
This PR adds the address validation function and other missing functions from the web version of native (mostly unimplemented but its better to have the stub).

This also exports the WebNative version for the default export on web, so the x-platform functionality works (partly)

many functions are unimplemented on web, and they will either be no-op or if `RadarRNWeb.throwOnUnimplemented(true);` is set, throw an error.

all the web apis are also exported via `RadarRNWeb` as well, so exisiting functionalities don't break. There were a few places where the argument processing in JS native needs to be added to fit the signature of the `radar-sdk-js` module. But existing functionalities are kept.

All calls that are implemented in web has been tested via the example app. Tested on Android + iOS as well to make sure they are still working as expected.


src/index.native.ts
```diff
@@ -274,15 +273,16 @@ const Radar: RadarNativeInterface = {
...
   setForegroundServiceOptions,
   setNotificationOptions,
-  acceptEvent,
-  rejectEvent,
   getTripOptions,
   startTrip,
-  updateTrip,
   completeTrip,
   cancelTrip,
+  updateTrip,
```
KennyHuRadar, line 283: do we need to make stubs for the mobile side of the sdk?

KennyHuRadar (approved):
This looks good to me

example/App.tsx
```diff
@@ -1,47 +1,46 @@
...
 
-Radar.on("log", (result) => {
+Radar.on("log", (result: string) => {
   console.log("log:", stringify(result));
 });
 
 export default function App() {
   // add in your test code here!
   const [displayText, setDisplayText] = useState("");
+  // const displayText = ""
```
KennyHuRadar, line 35: dead code

PR #362, opened 2024-10-30
CI should build android

ShiCheng-Lu:


lmeier:
Hmm, it didn't succeed, so re-running.

PR #367, opened 2024-11-26
add addRadarSDKVerify option

nickpatrick:
- expose `startVerifyServer()` and `stopVerifyServer()`, allowing customers to turn their Android app into a "companion app" (https://radar.com/documentation/fraud#web-and-desktop) for `trackVerified()` and `startTrackingVerified()` on mobile
- see also https://github.com/radarlabs/radar-sdk-android/pull/417 and https://github.com/radarlabs/radar-sdk-ios/pull/417
- adds `addRadarSDKVerify` plugin option to add `NanoHTTPD` dependency and `Telegraph` sub-dependencies (`CocoaAsyncSocket` and `HTTPParserC` with `:modular_headers => true`)

### DIFF
--- a/plugin/src/withRadarAndroid.ts
+++ b/plugin/src/withRadarAndroid.ts
@@ -59,6 +59,7 @@ export const withRadarAndroid = (
           <!-- for React Native -->
           <domain-config cleartextTrafficPermitted="true">
               <domain includeSubdomains="true">localhost</domain>
+              <domain includeSubdomains="true">10.0.2.2</domain>
           </domain-config>
       
           <!-- for SSL pinning -->
--- a/plugin/src/withRadarAndroid.ts
+++ b/plugin/src/withRadarAndroid.ts
@@ -83,7 +83,8 @@ export const withRadarAndroid = (
     if (config.modResults.language === "groovy") {
       config.modResults.contents = modifyAppBuildGradle(
         config.modResults.contents,
-        args.androidFraud ?? false
+        args.androidFraud ?? false,
+        args.addRadarSDKVerify ?? false
       );
     } else {
       throw new Error(
@@ -151,9 +152,10 @@ async function setCustomConfigAsync(
   return androidManifest;
 }
 
-function modifyAppBuildGradle(buildGradle: string, androidFraud: boolean) {
+function modifyAppBuildGradle(buildGradle: string, androidFraud: boolean, addRadarSDKVerify: boolean) {
   let hasLocationService = false;
   let hasPlayIntegrity = false;
+  let hasNanoHTTPD = false;
   if (
     buildGradle.includes(
       'com.google.android.gms:play-services-location:21.0.1"'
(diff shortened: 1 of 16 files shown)


plugin/src/withRadarAndroid.ts
```diff
@@ -59,6 +59,7 @@ export const withRadarAndroid = (
           <!-- for React Native -->
           <domain-config cleartextTrafficPermitted="true">
               <domain includeSubdomains="true">localhost</domain>
+              <domain includeSubdomains="true">10.0.2.2</domain>
```
nickpatrick, line 62: `localhost` on Android emulator

plugin/src/withRadarIOS.ts
```diff
@@ -74,6 +74,36 @@ export const withRadarIOS: ConfigPlugin<RadarPluginProps> = (config, args) => {
...
+          // Find the target block
+          const targetRegex = /target '(\w+)' do/g;
+          const match = targetRegex.exec(contents);
+          if (match) {
+            const targetStartIndex = match.index;
+            const targetEndIndex = contents.indexOf('end', targetStartIndex) + 3;
+
+            // Insert the pod declaration within the target block
+            const targetBlock = contents.substring(targetStartIndex, targetEndIndex);
+            const updatedTargetBlock = targetBlock.replace(/(target '(\w+)' do)/, `$1\n  pod 'RadarSDK/Verify', '3.19.2-beta.9'\n    pod 'CocoaAsyncSocket', :modular_headers => true\n  pod 'HTTPParserC', :modular_headers => true`);
```
nickpatrick, line 95: Telegraph sub-dependencies need `modular_headers => true`

plugin/src/withRadarIOS.ts
```diff
@@ -74,6 +74,36 @@ export const withRadarIOS: ConfigPlugin<RadarPluginProps> = (config, args) => {
...
   }
+  if (args.addRadarSDKVerify) {
+    config = withDangerousMod(config, [
+      'ios',
+      async config => {
+        const filePath = path.join(config.modRequest.platformProjectRoot, 'Podfile');
+        const contents = await fs.readFile(filePath, 'utf-8');
+
+        // Check if the pod declaration already exists
+        if (contents.indexOf("pod 'RadarSDK/Verify', '3.19.2-beta.9'") === -1) {
```
nickpatrick, line 85: @lmeier @KennyHuRadar Similar to `addRadarSDKMotion` option. Will we need to update the version by hand in both places?

ShiCheng-Lu: we should probably extract this to a constant in the future. Since we release all the sdks with matching versions (I think, since RadarSDKMotion releases a 3.19 together with a 3.19 main SDK even if there is no change). We can probably just update the version string in the podspec file, and we can read the version from there and use it across the plugin.

lmeier: Yea that's exactly correct — easiest to maintain the same version across the the main module and all submodules.

KennyHuRadar:
I noticed that you bumped the RN version, however, there are a few more spots where we need to bump the beta version and the native versions as well. It's documented [here](https://www.notion.so/radarlabs/How-to-release-the-React-Native-SDK-21c70f16e0bd4fe28425923645cadda0?pvs=4)

- we need to bump the rn version in the build.gradle
- we need to define the rn version in `android/src/main/java/io/radar/react/RNRadarModule.java` and `ios/RNRadar.m`
- we also need to bump ios native version in the cartfile

PR #373, opened 2024-12-18
[WIP] Spike on e2e testing

KennyHuRadar:
This PR attempts to do the following
- make example app more usable by e2e tester
    - have user inputed publishable key
- implement maestro flow to perform e2e
- configure building example app and running the e2e on EAS (not working yet)

Intention: We should be able to deduct that an app can build/run with react native radar and is able to call some important Radar calls, returning valid responses.


KennyHuRadar:
Decided to have e2e only run when manually triggered as eas build isn't cheap.
One limitation is that we also need to change the target of `react-naive-radar` from `file:..` to something that is available from the cloud. We will also need a script that can automatically perform a `npm i react-native-radar: BRANCH_FOR_TESTIN G`

PR #390, opened 2025-06-04
[WIP] RN migration

KennyHuRadar:
…eature


src/NativeRadar.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface Spec extends TurboModule {
+    // Core initialization
+    initialize(publishableKey: string, fraud?: boolean): void;
+
+    // Key tracking method
+    trackOnce(options?: Object): Promise<Object>;
+
+    // Event listener support
+    addListener(eventName: string): void;
```
KennyHuRadar, line 12: We want this to be compatible (interface wise) with the existing interface, lets change the method name to `on` and `off` per pervious interface.

src/index.ts
```diff
@@ -3,7 +3,8 @@ import { Platform } from "react-native";
 
 let nativeModule: RadarNativeInterface;
 
-nativeModule = Platform.OS === "web" ? require("./index.web").default : require("./index.native").default;
+// Use the new TurboModule architecture implementation
+nativeModule = Platform.OS === "web" ? require("./index.web").default : require("./index.turbo").default;
```
KennyHuRadar, line 7: nit: we will end up calling it something other than `./index.turbo`, but I too am not sure what we will call it. No need to address this at the moment, we will address it as we move towards converging this PR with the backwards compatibility stuff later on

android/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+cmake_minimum_required(VERSION 3.13)
```
KennyHuRadar, line 1: Is this file and `react-native-radar.cpp` generated by the codegen CLI? It is my impression from the docs that these files should not be committed and are autogenerated at build time. 
- https://github.com/reactwg/react-native-new-architecture/blob/main/docs/enable-libraries-android.md
- https://reactnative.dev/docs/the-new-architecture/using-codegen#android

KennyHuRadar:
branch abandoned, using different approach

PR #391, opened 2025-06-11
expose reason, transactionId, getProduct()

nickpatrick:


android/src/main/java/io/radar/react/RNRadarModule.java
```diff
@@ -310,8 +325,8 @@ public void onComplete(@NonNull Radar.RadarStatus status, @Nullable Location loc
     public void trackOnce(ReadableMap optionsMap, final Promise promise) {
         
         Location location = null;
-        RadarTrackingOptions.RadarTrackingOptionsDesiredAccuracy accuracyLevel = RadarTrackingOptions.RadarTrackingOptionsDesiredAccuracy.MEDIUM;
-        boolean beaconsTrackingOption = false;
+        RadarTrackingOptions.RadarTrackingOptionsDesiredAccuracy desiredAccuracy = RadarTrackingOptions.RadarTrackingOptionsDesiredAccuracy.MEDIUM;
+        boolean beacons = false;
```
lmeier, line 329: I support this renaming. @KennyHuRadar trying to remember why we were naming it like this? Was there some other boolean var we were getting confused by?

KennyHuRadar (approved):

PR #396, opened 2025-08-18
fix RN Map component

danielhausen:
the RN native Map component doesn't read styles in correctly as map options nor default to styles set in the dashboard

### DIFF
--- a/src/ui/map.jsx
+++ b/src/ui/map.jsx
@@ -105,9 +105,8 @@ const RadarMap = ({ mapOptions, children }) => {
         compassEnabled={false}
         logoEnabled={false}
         attributionEnabled
-        onRegionDidChange={mapOptions?.onRegionDidChange ? mapOptions.onRegionDidChange : null}
-        styleURL={styleURL}
-      >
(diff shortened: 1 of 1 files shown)


src/ui/map.jsx
```diff
@@ -105,9 +105,8 @@ const RadarMap = ({ mapOptions, children }) => {
         compassEnabled={false}
         logoEnabled={false}
         attributionEnabled
-        onRegionDidChange={mapOptions?.onRegionDidChange ? mapOptions.onRegionDidChange : null}
-        styleURL={styleURL}
-      >
+        onRegionDidChange={(mapOptions === null || mapOptions === void 0 ? void 0 : mapOptions.onRegionDidChange) ? mapOptions.onRegionDidChange : null} 
+        mapStyle={styleURL}>
```
jaspk06, line 109: for myself:
this was a breaking change for maplibregl rn v10 and was never migrated over

https://maplibre.org/maplibre-react-native/docs/setup/migrations/v10/#mapview-props-styleurl-and-stylejson-unified-to-mapstyle

jaspk06 (approved):

PR #405, opened 2025-09-11
feat: add MapLibreGL loading events to Radar Map component

danielhausen:
- Add onDidFinishLoadingMap event forwarding
- Add onWillStartLoadingMap event forwarding
- Add onDidFailLoadingMap event forwarding
- Update TypeScript definitions for new events
- Maintains existing mapOptions API pattern
- Fixes issue where loading events were ignored

Resolves customer request for onDidFinishLoadingMap support

**New Architecture**
new-arch-android

new-arch-ios

**Old Architecture**
old-arch-ios

old-arch-android

### DIFF
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -31,7 +31,7 @@ if (host) {
 }
 
 
-Radar.initialize("prj_test_pk_e3bd55ab250955926cdd76c2baeb2ed559434925", true);
+Radar.initialize("prj_test_pk_", true);
 const stringify = (obj: any) => JSON.stringify(obj, null, 2);
 declare global {
   var __turboModuleProxy: any;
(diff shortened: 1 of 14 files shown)


example/App.tsx
```diff
@@ -30,7 +31,7 @@ if (host) {
 }
 
 
-Radar.initialize("prj_test_pk_", true);
+Radar.initialize("prj_test_pk_e3bd55ab250955926cdd76c2baeb2ed559434925", true);
```
KennyHuRadar, line 34: ops, might want to rotate your key

KennyHuRadar:
Just as a sanity check we tried running this on both platform, on both architecture?

PR #440, opened 2026-04-27
FENCE-2791: Add plugin for `autoHandleNotificationDeeplink` on ios

alanjcharles14:
## Summary
Adds Expo config plugin support for iOS notification deep link and conversion logging options. Previously, `autoHandleNotificationDeepLinks` and `autoLogNotificationConversions` could only be set via the JS `Radar.initialize(...)` call, which runs after the React Native bundle has loaded — too late for iOS to deliver cold-start notification taps to the SDK's `UNUserNotificationCenterDelegate` swizzle. As a result, deep linking from a tapped notification on a killed app would foreground the app to its default route instead of routing to the linked URL.

## New SDK Components

**`iosAutoHandleNotificationDeepLinks`** — New iOS-only `RadarPluginProps` boolean. When `true`, the plugin injects a `RadarInitializeOptions` instance into the app's `AppDelegate` with `autoHandleNotificationDeepLinks = YES` and calls `[Radar nativeSetup:options]` after the `super` call in `didFinishLaunchingWithOptions`. This installs Radar's notification-delegate swizzle early enough in app launch to handle cold-start deep links, fixing the case where tapping a Radar campaign notification on a killed app would land the user on the home screen instead of the linked URL.

**`iosAutoLogNotificationConversions`** — New iOS-only `RadarPluginProps` boolean. When `true`, the plugin sets `autoLogNotificationConversions = YES` on the same `RadarInitializeOptions` instance passed to `[Radar nativeSetup:options]`, enabling automatic association of `opened_app` conversions with the notification that opened the app — without requiring users to manually edit their `AppDelegate` or `UNUserNotificationCenterDelegate` implementation.

**`withRadarAppDelegate`** — New internal config-plugin mod in `plugin/src/withRadarIOS.ts` that wraps Expo's `withAppDelegate`. It detects the AppDelegate language (`objc` / `objcpp` / `swift`), inserts the appropriate `RadarSDK` import, and injects a language-appropriate `nativeSetup` block immediately after `didFinishLaunchingWithOptions`'s `super` call. Idempotent via a sentinel comment, gated on at least one of the new props being enabled, and falls back to `WarningAggregator.addWarningIOS` (rather than throwing) if it can't find the expected anchor or encounters an unsupported AppDelegate language.


example/package-lock.json
```diff
@@ -6847,31 +6840,8 @@
       }
     },
     "node_modules/react-native-radar": {
```
stevepopovich, line 6849: So this example app is an Expo app? And this diff would be expected for folks that upgrade to the new version of this SDK?

plugin/src/withRadarIOS.ts
```diff
@@ -1,9 +1,115 @@
...
+  WarningAggregator,
+} = require("expo/config-plugins");
 import type { RadarPluginProps } from "./types";
 const fs = require('fs/promises');
 const path = require('path');
 const pkg = require("../../package.json");
 
+const NATIVE_SETUP_SENTINEL = "@generated react-native-radar nativeSetup";
+
+function buildObjcSnippet(args: RadarPluginProps): string {
```
stevepopovich: Is it possible to write unit tests for this, or should we? Writing swift and Obj C manually in a typescript file seems scary to me, but very new to this

plugin/src/types.ts
```diff
@@ -9,4 +9,18 @@ export interface RadarPluginProps {
     androidActivityRecognition?: boolean;
     addRadarSDKMotion?: boolean;
     iosNSMotionUsageDescription?: string;
+
+    /**
+    * iOS only. When true, injects `[Radar nativeSetup:options]` into AppDelegate
+    * with autoHandleNotificationDeepLinks = YES so cold-start notification taps
+    * route to the URL in the notification payload.
+    */
+    iosAutoHandleNotificationDeepLinks?: boolean;
```
stevepopovich, line 18: Should we/is it possible to default this behavior to true? Seems like something most users would expect to work out of the box

stevepopovich (commented):
Asking questions as a total noob!

example/package-lock.json
alanjcharles14: to be honest I'm not entirely sure what happened here. I know we changed some of our npm/node requirements lately but this doesn't seem correct. The stuff that got removed probably should never have been there? We want to test the package directly from the SDK not from npm itself. So, no a customer shouldn't expect to see this because they shouldn't be linking it locally. Happy to follow up on this offline for a better explanation!

plugin/src/withRadarIOS.ts
alanjcharles14: we should in general do a lot better than we currently do in this repo wrt testing in general. It's something on my list of things to accomplish. However, this is a bit difficult to unit test "correctly" because it's actually writing these and injecting them into AppDelegate at build-time. Where it actually injects those is the important piece. A scary, yet standard practice for Expo apps since you don't control the native files. Another thing I'm happy to cover offline!

plugin/src/types.ts
alanjcharles14: nah I don't think so. There's a fair amount of set-up involved with deep-links and push notifications in general. Since we're modifying native file functionality i only want customers to do it when they need it.

stevepopovich: Makes sense 👍🏼

plugin/src/withRadarIOS.ts
stevepopovich: Thats fine! I need to learn a bit more about it

example/package-lock.json
stevepopovich: Seems fine for now...mostly just asking. I am happy to assume package-lock.json is a block box we don't need to look at it too closely

stevepopovich (dismissed):

lmeier (dismissed):

lmeier (approved):

Issue #442, opened 2026-05-19
trackVerified() returns ERROR_PLUGIN on iOS after upgrading from 3.20.4 to 4.31.0 (Expo)


florian-deroo:
### Description

After upgrading `react-native-radar` from **3.20.4 → 4.31.0**, `Radar.trackVerified()` consistently fails on iOS with `ERROR_PLUGIN`. The exact same code and Expo plugin configuration worked without any issue on 3.20.4.

### Error

```js
{
  "errorType": "acquisition",
  "message": "An unexpected error occurred while checking your location. Please try again.",
  "originalError": [Error: ERROR_PLUGIN],
  "radarErrorType": "UNKNOWN",
  "success": false,
  "title": "Location Verification Failed"
}
```

This is **reproducible 100% of the time**. The user has granted foreground location permissions, and other Radar APIs (`Radar.getPermissionsStatus()`, `Radar.setUserId()`, etc.) work fine — only `trackVerified()` fails.

---

### What I think might be happening

`ERROR_PLUGIN` seems to indicate that an iOS-side fraud/verification submodule isn't being loaded at runtime when `trackVerified()` is called. Since I kept the exact same Expo plugin config (`iosFraud: true`) and the same app bundle identifier across the upgrade, my guess is that something in the iOS native packaging or the Expo plugin changed between 3.20.4 and 4.x regarding how the fraud module is bundled — but I might be wrong. Happy to provide more debug logs if useful.

---

### Environment

| Package / Setting | Version |
|---|---|
| `react-native-radar` | **4.31.0** (previously **3.20.4** — working) |
| `expo` | 55.0.13 |
| `react-native` | 0.83.4 |
| `react` | 19.2.0 |
| Platform | iOS (build via EAS) |
| New Architecture | enabled (`newArchEnabled: true`) |

---

### Expo plugin configuration

> Unchanged across the upgrade. We are an iOS app only.

```json
[
  "react-native-radar",
  {
    "iosFraud": true,
    "iosBackgroundMode": false,
    "androidFraud": false,
    "androidBackgroundPermission": false,
    "androidFineLocationPermission": true,
    "addRadarSDKMotion": false,
    "iosNSLocationWhenInUseUsageDescription": "X.",
    "iosNSLocationAlwaysAndWhenInUseUsageDescription": "X"
  }
]
```

---

### Reproduction

```ts
Radar.initialize(process.env.EXPO_PUBLIC_RADAR_TOKEN, true);
[...]

PR #443, opened 2026-05-19
Fraud plugin

alanjcharles14:
## Summary

Fixes `Radar.trackVerified()` failing with `ERROR_PLUGIN` on iOS in 4.x. The iOS Radar SDK split fraud/attestation into a separate `RadarSDKFraud` module around v3.27, but `react-native-radar` 4.x only vendored `RadarSDK` and `RadarSDKMotion` — leaving `NSClassFromString(@"RadarSDKFraud")` to return `nil` at runtime for any consumer calling `trackVerified()`. This PR vendors the fraud xcframework, expands `iosFraud: true` to actually wire it into the build, and gives a clearer error message if the module ever ends up unlinked.

## New SDK Components

**`RadarSDKFraud.xcframework`** — Vendored binary from [`radarlabs/radar-sdk-ios-fraud-spm@1.2.0`](https://github.com/radarlabs/radar-sdk-ios-fraud-spm/releases/tag/1.2.0), pinned by SHA256, committed to git alongside the existing `RadarSDK` and `RadarSDKMotion` xcframeworks.

**`RadarSDKFraud.podspec`** — New opt-in pod that vendors the xcframework, depends on the main `Radar` pod, and is gated behind the existing `iosFraud: true` Expo plugin flag (no new config knobs).

**`Radar.trackVerified` runtime guard** — `RNRadar.mm` now pre-flights `NSClassFromString(@"RadarSDKFraud")` and rejects with an actionable message pointing the user at `iosFraud: true` instead of a bare `ERROR_PLUGIN`.

## iOS

**Before**

image

**After**
image

## Android

**Before**
image

**After**
image


MIGRATION.md
```diff
@@ -43,6 +43,40 @@ const mapRef = useRef<MapRef>(null);
 <RadarMap mapOptions={{ mapRef }} />
 ```
 
+### Breaking: `Radar.trackVerified()` now requires opting in to the fraud module
```
stevepopovich, line 46: Should this be in the main README? Not strictly a migration but I could be wrong.

stevepopovich (approved):
Really impressive work! Doesn't look like the most fun day at the office lol.

The only thing that feels "smelly" is all the generated files for iOS. Are we sure they should all be checked in? Probably yes, but just double checking. I am SUPER unfamiliar with this type of work so...

MIGRATION.md
alanjcharles14: erm- i think there's reason enough to argue it should be added into our main documentation page but we don't really include consumer-based implementation stuff in the readme so I think it's okay here. To be honest, users already basically had to do this before the fix (`iosFraud: true` in `app.json`) but figured i'd call it out here since the stuff happening under the hood is a bit different.

alanjcharles14:
@stevepopovich yaaaa it's definitely smelly. Yes, I think they should be checked in so we don't need to download all of the xcFrameworks every time someone pulls down the repo. I think it would require a fair amount of other work to make sure these are included in the bundled npm package in our release action if they weren't checked in. That honestly might be worth doing but this is a short-term-ish bandaid until we can finally just move to pulling directly from SPM (which react native does not yet support doing).

MIGRATION.md
stevepopovich: thats fair! 👍🏼

lmeier (approved):
Incredible effort.

alanjcharles14:
resolves #442

PR #444, opened 2026-06-11
Alancharles/fence 2880 react native ota updates improvements

alanjcharles14:
Fixes a native crash (iOS `SIGSEGV`/`SEGV_ACCERR`, Android native heap corruption) that can occur on the new architecture when the SDK emits a location/event callback into the react native runtime while that runtime is being torn down (e.g. during an OTA update reload).


nickdonnelly (approved):
lgtm

PR #446, opened 2026-06-24
FENCE-2893: Send `X-Radar-Mobile-Origin` on `RadarMap` requests to support mobile app restrictions

alanjcharles14:
## Summary

`RadarMap` tiles/styles are fetched directly by MapLibre, bypassing the native SDK's networking. As a result, those requests never carried the iOS bundle ID / Android package name that mobile app restrictions validate against. When a project had **both** website restrictions and mobile restrictions enabled, the maps API required a request to match an allowed web origin *or* mobile ID — and the MapLibre requests sent neither, so they were rejected (401) and the map failed to load. Normal SDK calls (e.g. tracking) were unaffected because they already send `X-Radar-Mobile-Origin`.

This PR makes `RadarMap` attach `X-Radar-Mobile-Origin` (the app's bundle ID / package name) to all Radar maps requests so they satisfy mobile restrictions.

## Changes

- Added a native `getMobileOrigin()` method that returns the iOS bundle identifier / Android package name. The same value the native SDK sends as `X-Radar-Mobile-Origin`. 
- `map.jsx`: register an `X-Radar-Mobile-Origin` header via MapLibre's `TransformRequestManager.addHeader`, scoped (via `match`) to the Radar maps host so it's only sent to `api.radar.io/maps/` and not to any third-party tile sources.

## Testing

- **Server contract (curl):** against a live project with both website + mobile restrictions:
  - no header → `401`
  - `-H "X-Radar-Mobile-Origin: <bundleId>"` → `200`
- **On-device (iOS + Android):** example app with the restricted live key:
  - both restrictions + correct bundle ID/package name → map loads
  - incorrect/removed mobile ID → map fails (confirms enforcement, not bypass)
  - no restrictions → map loads (regression check)
  - tracking continues to work

### DIFF
--- a/src/ui/map.jsx
+++ b/src/ui/map.jsx
@@ -1,23 +1,26 @@
 import React, { useState, useEffect } from 'react';
 import { View, Image } from 'react-native';
 import Radar from '../index.native';
-import { getHost, getPublishableKey } from '../helpers';
+import { getHost, getPublishableKey, getMobileOrigin } from '../helpers';
 import styles from './styles';
 
 let MapLibreGL;
 let MapLibreMap;
 let GeoJSONSource;
 let Layer;
+let TransformRequestManager;
 try {
   MapLibreGL = require('@maplibre/maplibre-react-native');
   MapLibreMap = MapLibreGL.Map;
   GeoJSONSource = MapLibreGL.GeoJSONSource;
   Layer = MapLibreGL.Layer;
+  TransformRequestManager = MapLibreGL.TransformRequestManager;
 } catch (e) {
   MapLibreGL = null;
   MapLibreMap = null;
   GeoJSONSource = null;
   Layer = null;
+  TransformRequestManager = null;
 }
 
 const DEFAULT_STYLE = 'radar-default-v1';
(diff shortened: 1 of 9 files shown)


src/ui/map.jsx
```diff
@@ -49,10 +52,35 @@ const RadarMap = ({ mapOptions, children }) => {
...
+      const host = await getHost();
+
+      if (TransformRequestManager) {
+        try {
+          const mobileOrigin = await getMobileOrigin();
+          if (mobileOrigin) {
+            const mapHostPattern = `${host
+              .replace(/^https?:\/\//, '')
+              .replace(/\./g, '\\.')}/maps/`;
+            TransformRequestManager.addHeader({
```
nickdonnelly, line 65: I'm curious if it makes sense to maintain a fork of MapLibre that handles this under the hood.

nickdonnelly (approved):

src/ui/map.jsx
alanjcharles14: Hmm, I definitely don't want to maintain a fork if we can at all help it, but we have run into other issues with this dependency in the past so I'm not inherently against it. Why would we want to manage this under the hood exactly? It's not really exposed to the customer as-is and is really just a limitation with our `Mobile Restrictions` functionality.

nickdonnelly: My train of thought was geared towards a future where we have more bespoke map features and we need to reach into MapLibre to make them happen. Not advocating for this in this trivial case but it's in the back of my mind as we start ramp up more work on maps.

PR #450, opened 2026-09-04
[FENCE-3001] Automate iOS SDK propagation

stevepopovich:
## Summary

- Adds a separate `update-ios-sdk.yml` workflow for manually triggered, best-effort iOS-to-React Native propagation. The existing `update-native-sdks.yml` workflow is unchanged.
- Resolves GitHub's latest stable Radar iOS release when an engineer selects **Run workflow**.
- Ignores equal or older releases, calculates the next React Native minor version, rebuilds static RadarSDK and RadarSDKMotion frameworks, and leaves Fraud unchanged.
- Gives Claude one bounded attempt to inspect the latest iOS source and pinned Android source and port useful bridge, TypeScript, test, mock, and Example app changes.
- Runs focused Node validation once and still creates or updates one draft PR when Claude or validation fails, labeling every generated draft `ios-sdk-update` and `needs-human` for human review.
- Uses the workflow's built-in `GITHUB_TOKEN` to publish the draft PR and pins action dependencies to commit SHAs; no GitHub App is required for this manual phase.

## Setup

- Add `ANTHROPIC_API_KEY` as an organization secret scoped to this repository.
- The repository already allows GitHub Actions to create pull requests with its built-in token.
- The `ios-sdk-update` and `needs-human` labels are pre-created in the repository.

No package is published or merged automatically.


.github/workflows/update-ios-sdk.yml
```diff
@@ -0,0 +1,161 @@
...
+            ## Validation
+
+            - Claude step: `${{ steps.claude.outcome }}`
+            - Node, TypeScript, and Jest validation: `${{ steps.validation.outcome }}`
+            - [Workflow run](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
+            - CircleCI Node, iOS, and Android jobs provide independent validation on this draft.
+          labels: |
+            ios-sdk-update
+            needs-human
+          add-paths: |
```
alanjcharles14: Claude can modify files outside this add-paths list. Those changes are included during validation but dropped from the PR. Can we either include all changed files or fail when a changed file isn’t listed?

stevepopovich: ok yeah, I think we can drop this and include everything, fixed. Good catch!

alanjcharles14 (approved):

